PR #179, opened 2014-12-15
Set the Forwarded header for POST and DELETE calls.

akrabat:


lornajane:
What do we do with these settings once we've set them.  Where are they used?

akrabat:
POST goes to the spam filter. DELETE is for consistency.

PR #198, opened 2015-02-21
Implement SearchController::getCache()

akrabat:


lornajane:
This is to fix an issue seen when you try to use the search feature, it produces errors on dev, nothing on live

PR #207, opened 2015-02-28
Jira 482 password change

mgdm:
Updated @cyberspice's branch with latest master and confirmed it works. It originally made what is now the profile link at the top of the page point to a password reset. I've changed this so it's another item that appears when logged in. This might need reviewed.


akrabat:
Waiting on [API PR 157](https://github.com/joindin/joindin-api/pull/157)

lornajane:
API PR replaced, now https://github.com/joindin/joindin-api/pull/168 BUT it will require changes here because the API endpoint has changed.

Consumers (trusted consumers only) should now PUT to {user_uri}/password with fields "old_password" and "password" in the JSON body.

akrabat:
Closing as I want to do edit profile within the same page. When the endpoint is in, I'll look at this code to see what we can use and either copy over or use this as the base.

PR #155, opened 2014-09-28
Adds possibility to edit an event in web2

heiglandreas:
This commit adds the possibility to edit an event via the web2 interface

PR #210, opened 2015-03-06
Alow editing events in web2

heiglandreas:
This PR should enable editing events in the web2-interface. 

It is the successor of PR #155 which this one obsoletes!

When you are allowed to edit an event either because you are one of the event-hosts or you are an admin you find an edit-button in the event-header. Clicking will bring you to a form where you can edit all the values of the event similar to the form you used to create the event.

On submitting the form-data is stored via the joindin API using the PUT method.


akrabat:
Sorry, but my first edit test caused this:

```
Type: Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException
Message: Neither the property "href" nor one of the methods "addHref()"/"removeHref()", "setHref()", "href()", "__set()" or "__call()" exist and have public access in class "Event\EventEntity".
File: /vagrant/joindin-web2/vendor/symfony/property-access/Symfony/Component/PropertyAccess/PropertyAccessor.php
Line: 456
```

PR #211, opened 2015-03-11
Changes daterange-lib to current version

heiglandreas:
This PR updates the DateRange-lib to the current version. That should also display the daterange for a one-day event correctly.


akrabat:
You don't ned to update the composer.son in order for us to pick up 1.1.

However, as we store the composer dependencies in our repository, you need to do this:

    composer update org_heigl/daterange

heiglandreas:
So I check in the ```composer.lock```? I normally try to avoid that as my one might be faulty :wink:

akrabat:
Yes. In general `composer.lock` should be committed. We're a little unusual in that we also commit all of `vendor`.

PR #218, opened 2015-03-25
User profile edit

akrabat:
Enable a user to edit their name, email, Twitter username and change their password.

Note that this PR depends on API PRs: [169](https://github.com/joindin/joindin-api/pull/169) & [165](https://github.com/joindin/joindin-api/pull/165) which have been merged.

This will close: [JOINDIN-482](https://joindin.jira.com/browse/JOINDIN-482) and [JOINDIN-554](https://joindin.jira.com/browse/JOINDIN-554)

### DIFF
--- a/app/templates/User/profile.html.twig
+++ b/app/templates/User/profile.html.twig
@@ -16,12 +16,23 @@
                         Twitter: <a href="https://twitter.com/{{ thisUser.getTwitterUsername }}">
                             @{{ thisUser.getTwitterUsername }}</a>
                         {% endif %}
+
+                        <p class="more">
+                            <a href="{{ urlFor('user-profile-edit', {'username': thisUser.getUsername}) }}">
+                                Edit</a>
+                        </p>
                     </section>
                 </div>
             </div>
         </div>
     </section>
 
+    {% if flash.getMessages.message %}
+        <div class="alert alert-success">{{flash.getMessages.message}}</div>
+    {% endif %}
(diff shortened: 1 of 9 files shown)


app/templates/User/profile.html.twig
```diff
@@ -16,12 +16,23 @@
                         Twitter: <a href="https://twitter.com/{{ thisUser.getTwitterUsername }}">
                             @{{ thisUser.getTwitterUsername }}</a>
                         {% endif %}
+
+                        <p class="more">
+                            <a href="{{ urlFor('user-profile-edit', {'username': thisUser.getUsername}) }}">
+                                Edit</a>
```
lornajane: Can we only show this if the user actually can edit?

lornajane:
When a site admin edits a user, they can't change the user's password so those fields shouldn't be visible.

Also it isn't possible to remove a twitter account, this is valid as an empty string but it doesn't save when the field is empty - the API will tolerate an empty string but this is a null value.

akrabat:
To fix the null issue on twitter_username, I have created API [PR 176](https://github.com/joindin/joindin-api/pull/176).

Looking at the code, the conversion of an empty string to null is hardcoded into the Symfony2 Form component's `viewToNorm` method. The 'correct' solution is to write a "view transformer" class that would allow us to bypass the hardcoded conversion. If we don't go for the API change, then I'll create one.

PR #219, opened 2015-03-27
JOINDIN-564 Logout should leave me on the same page

kazu9su:
Hi team,
This is my first contribution of this project.
I tried to solve below issue.
https://joindin.jira.com/browse/JOINDIN-564

Thanks!
Kazu

### DIFF
--- a/app/src/Search/SearchController.php
+++ b/app/src/Search/SearchController.php
@@ -87,7 +87,6 @@ class SearchController extends BaseController
                 'page'     => $page,
                 'keyword'  => $keyword,
                 'tag'      => $tag,
-                'redirect' => $this->getPath(),
             )
         );
     }
@@ -130,7 +129,6 @@ class SearchController extends BaseController
                 'page'       => $page,
                 'pagination' => $pagination,
                 'keyword'    => $keyword,
-                'redirect'   => $this->getPath(),
             )
         );
     }
--- a/app/src/Search/SearchController.php
+++ b/app/src/Search/SearchController.php
@@ -83,10 +83,11 @@ class SearchController extends BaseController
         $this->render(
             'Event/search.html.twig',
             array(
-                'events'  => $events,
-                'page'    => $page,
-                'keyword' => $keyword,
-                'tag'     => $tag
+                'events'   => $events,
+                'page'     => $page,
+                'keyword'  => $keyword,
+                'tag'      => $tag,
+                'redirect' => $this->getPath(),
             )
         );
     }
@@ -128,7 +129,8 @@ class SearchController extends BaseController
                 'talks'      => $talks,
                 'page'       => $page,
                 'pagination' => $pagination,
-                'keyword'    => $keyword
+                'keyword'    => $keyword,
+                'redirect'   => $this->getPath(),
             )
         );
     }
--- a/app/src/Search/SearchController.php
+++ b/app/src/Search/SearchController.php
@@ -83,10 +83,10 @@ class SearchController extends BaseController
         $this->render(
             'Event/search.html.twig',
             array(
-                'events'   => $events,
-                'page'     => $page,
-                'keyword'  => $keyword,
-                'tag'      => $tag,
+                'events'  => $events,
+                'page'    => $page,
+                'keyword' => $keyword,
+                'tag'     => $tag
             )
         );
     }
@@ -128,7 +128,7 @@ class SearchController extends BaseController
                 'talks'      => $talks,
                 'page'       => $page,
                 'pagination' => $pagination,
-                'keyword'    => $keyword,
+                'keyword'    => $keyword
             )
         );
     }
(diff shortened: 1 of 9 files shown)


akrabat:
ok to test

lornajane:
The build for this pull request has failed, you can view the output of the build (http://jenkins.joind.in/job/web2-pr-build/255/console) to find the exact error messages, but they are all coding standards fixes.  We use PSR2 so please correct the issues (mostly whitespace) and when the build passes, we'll take another look at the patch.

kazu9su:
ok I'll check it and fix coding style.
The other hand,  I found an error in self-test.
So, I'll fix it too and send new pull request.
Thanks!

akrabat:
There's no need to close this PR. You can make the changes on your `JOINDIN-564` branch and push them. Github will notice and update this PR appropriately.

kazu9su:
Thanks for your advice, Rob:)
So please wait a moment to push code...

kazu9su:
I pushed my changes to JOINDIN-564 branch.
Please confirm that.

Thanks!

akrabat:
ok to test

akrabat:
One error left: http://jenkins.joind.in/job/web2-pr-build/258/console

kazu9su:
I pulled changes again

akrabat:
I think that passing a `redirect` variable to the view in every single controller action isn't the right approach. 

It would be better to either add a variable called `currentPath` to the [Twig global context](http://twig.sensiolabs.org/doc/advanced.html#globals) or by using a [Twig extension function](http://twig.sensiolabs.org/doc/advanced.html#functions) accessed as `{{ currentPath() }}`.

Also, in `layout.html.twig`, I don't want to see  `?redirect=` when there isn't a path set be set.

kazu9su:
Thank you for your review.
OK, I'll refactoring my code

kazu9su:
I pushed all my changes.
Please confirm it again.
Thanks!

app/src/Search/SearchController.php
```diff
@@ -83,10 +83,10 @@ public function searchEvents()
...
             'Event/search.html.twig',
             array(
-                'events'  => $events,
-                'page'    => $page,
-                'keyword' => $keyword,
-                'tag'     => $tag
+                'events'   => $events,
+                'page'     => $page,
+                'keyword'  => $keyword,
+                'tag'      => $tag,
```
heiglandreas: Why the extra space?

app/src/User/UserController.php
```diff
@@ -122,8 +124,8 @@ public function register()
         $this->render(
             'User/register.html.twig',
             array(
-                'form' => $form->createView(),
-            )
+                'form'     => $form->createView(),
```
heiglandreas: Why the extra space?

app/src/User/UserController.php
```diff
@@ -604,7 +609,7 @@ public function remindUsername()
         $this->render(
             'User/username-reminder.html.twig',
             array(
-                'form' => $form->createView(),
+                'form'     => $form->createView(),
```
heiglandreas: And again: why the extra space? :wink:

akrabat:
I think we're definitely on the right track now. 

Next things:

1. Please revert all unnecessary changes in the controllers, including the change to the Login link as this issue is for log out.
2. It doesn't actually work. When I click log out, I always end up on the home page.

kazu9su:
OK so at first, I revert unnecessary changes.

kazu9su:
and about 2nd problem.
I cannot confirm( or sorry but cannot understand..) this problem.
Could you explain in detail?

Thanks a lot!

heiglandreas:
I can reproduce what @akrabat describes. I open up the joind.in-website (web2) and go to any events detail-page. I then click onto the login-button and login. I'd expect that, after a successfull login, I'd be redirected to the events detail-page I've been on before. Instead I see the "Hot Events" and the URL reads ```<server-address>:<port>```

The same happens when I'm on any detail-page and click "logout". I'd expect to see the detail-page again, but instead I see the "Hot Events". The URL reads ```<server-address>:<port>/?redirect=/event/ruby-meetup```.

akrabat:
Note that this issue & PR should only be for handling the logout link.

kazu9su:
Thank you all.

Okay, I'll fix it!

kazu9su:
Hi all,
I've fixed the problem.
Please confirm it.

By the way, there is something I'm worrying about a little.
The route "/user/<username>" should have a login filter.
How about it?

Thanks all!

akrabat:
Thanks kazu9su, I'll review the changes when I can.

/user/username certainly shouldn't have a login guard as it shows useful public information about a given user. e.g. https://m.joind.in/user/akrabat

kazu9su:
Hi akrabat,
I see.
Thanks for your cooperation and kindness!

akrabat:
Committed to master and now live!

Thank you @kazu9su :)

PR #221, opened 2015-04-03
Update Symonfy components to 2.6.6

akrabat:
This fixes a bug in the repeated form element where from 2.6.0 to 2.6.4,
it didn't actually work when the data submitted in the two elements did
not match.

This is the bug: symfony/symfony#12792
This is the fix: symfony/symfony#13198


akrabat:
This fixes a bug in the repeated form element where from 2.6.0 to 2.6.4, it didn't actually work when the data submitted in the two elements did not match.

This is the bug: https://github.com/symfony/symfony/issues/12792
This is the fix: https://github.com/symfony/symfony/pull/13198

PR #223, opened 2015-04-09
Event page tidy up

akrabat:
Rework the event page:

* Header contains just the key information
* Menu under header
* Data about the event is now on the details page rather than mixed between header and sidebar
* Map page no longer separate - now on details page
* Event comments are a separate page

Looks like this:
Event detail page


lornajane:
Can you revisit this?  It won't cleanly merge because it's wrapped around the event ratings and I'm not sure how to resolve the conflict because I think some of the code in the other pull request needs moving to somewhere new in this one.  Branch looks good though, I really like the changes, I just couldn't unpick the merge.

akrabat:
Rebased and fixed event comments to handle the new rating element.

PR #227, opened 2015-04-26
Frontend for forgot password

lornajane:
This adds a "forgot password" link to the login page, which prompts the user for their username (there is already a forgot username feature).  This sends them an email with a link in it.

This feature also has the landing page for the link; the user is prompted to pick a new password, and then asked to log in once it has been successfully set.


lornajane:
This relies on https://github.com/joindin/joindin-api/pull/185

app/src/User/UsernameInputFormType.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace User;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * Form used to accept input of an email address only
```
akrabat: Comment should say "of a username only" - I'll update on merge, but putting here as a reminder to myself!

PR #228, opened 2015-05-05
Only show map if co-ordinates exist and are not 0

railto:
Ensures that map will only display if co-ordinates have been set and are not 0

Fixes #JOINDIN-560


akrabat:
ok to test

PR #232, opened 2015-06-07
Add a gap between the user's gravatar and the heading border

akrabat:


lornajane:
I guessed from the branch name this was for the user profile page and couldn't spot it anywhere else so I merged it!

PR #235, opened 2015-06-20
Approve/Reject pending events

akrabat:
Depends on API https://github.com/joindin/joindin-api/pull/189

### DIFF
--- a/app/src/Event/EventApi.php
+++ b/app/src/Event/EventApi.php
@@ -325,4 +325,36 @@ class EventApi extends BaseApi
 
         return $commentData;
     }
+
+    /**
+     * Approve a pending event by POSTing to approval_uri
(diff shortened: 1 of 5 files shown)


lornajane:
Little bit of weirdness on this one, when I click either approve or reject, it goes to http://web2.dev.joind.in/event/action-pending-event and I get page not found and the events aren't approved or rejected.

I see this one has the cream bar, is that going to hold more links or specifically admin tasks?

akrabat:
I can't reproduce (yet, at least). From your description, it sounds like the event doesn't have a friendly name, but I haven't worked out how that can happen.

General plan is for the cream bar to always only contain admin tasks. If we need more links on the event list page, then that'll be a separate control.

PR #241, opened 2015-08-24
 JOINDIN-535 Added CFP details on events detail page

shaunmza:
Added new helper function to display date time according to a date format,
added function in the events entity file to return the current CFP status based on start and end dates
finally, changes to the template to display CFP details

### DIFF
--- a/app/templates/Event/details.html.twig
+++ b/app/templates/Event/details.html.twig
@@ -43,6 +43,19 @@
                     Event Website: <a href="{{ event.getWebsiteAddress }}" rel="nofollow">{{ event.getWebsiteAddress }}</a> <i class="fa fa-external-link-square"></i>
                     <br>
                     Short URL: <a href="{{ shortUrlForEvent(event.getStub) }}">{{ shortUrlForEvent(event.getStub) }}</a> <i class="fa fa-external-link-square"></i>
+                    {% if event.getCfpStatus %}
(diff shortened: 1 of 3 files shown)


akrabat:
ok to test

app/templates/Event/details.html.twig
```diff
@@ -43,6 +43,19 @@
                     Event Website: <a href="{{ event.getWebsiteAddress }}" rel="nofollow">{{ event.getWebsiteAddress }}</a> <i class="fa fa-external-link-square"></i>
                     <br>
                     Short URL: <a href="{{ shortUrlForEvent(event.getStub) }}">{{ shortUrlForEvent(event.getStub) }}</a> <i class="fa fa-external-link-square"></i>
+                    {% if event.getCfpStatus %}
+                        <br>
+                        Call for Papers Status: {{ event.getCfpStatus }}
+                    {% endif %}
+                    {% if event.getCfpStatus == 'Open' %}
+                        <br>
+                        (ends {{ dateFormat(event.getCfPEndDate, 'd.m.Y') }})
```
akrabat: Twig has a [date filter](http://twig.sensiolabs.org/doc/filters/date.html) which would be more appropriate to use if it works with DateTime objects.

akrabat:
Closed in favour of #248

PR #245, opened 2015-09-06
JOINDIN-493 - Label for Average talk rating at top right of talk page

dstockto:


app/templates/Talk/_common/talk_header.html.twig
```diff
@@ -39,6 +39,7 @@
                 </div>
                 <div class="col-xs-4 text-right">
                     <div style="margin-top: 10px">
+                        Avg. Rating
```
akrabat: This label should not display if there is no rating.

dstockto: I'll update and send it back up.

Thank you

dstockto:
@akrabat Updated to only show if there is a rating to show

PR #247, opened 2015-09-07
Config to make all API calls go through a proxy

lornajane:
When debugging, I find it helpful to push my API calls through a proxy such as Charles or mitmproxy, but it needs a minor modification to the code that makes the API calls to add the proxy setting to the stream context.  This patch makes this configurable and doesn't proxy if the setting is missing or empty.


akrabat:
I really like this ability! Makes it much easier to see what's happening.

PR #248, opened 2015-09-08
Joindin 535 Add CFP details onto events detail page 

shaunmza:
Removed custom function and changed the template to use twig date filter


akrabat:
ok to test

PR #254, opened 2015-09-26
JOINDIN-594: Full description for pending events

dannym87:
Fixing event description being truncated on pending events list


lornajane:
ok to test

PR #255, opened 2015-09-26
JOINDIN-601 Sort the tracks in the Schedule.

waterada:
See: https://joindin.jira.com/browse/JOINDIN-601

Now, the track names in the schedule aren't sorted at all. This is the thing to sort them.


lornajane:
ok to test

lornajane:
I like this idea but can we regroup on the implementation?
 - I'd rather see a sort order column added than this alphabetical sort
 - this functionality should be in the API so that the apps and js widgets (and any other consumers) can be consistent and use the same feature to display the information the same way if they want to

PR #256, opened 2015-09-27
Localize only talk comment page

waterada:
JOINDIN-603 Localization of the only Talk Comment page
See: https://joindin.jira.com/browse/JOINDIN-603


yandod:
If we could publish this change by Friday, it would be really nice for the attendees of PHP Conference Japan.
We will have 2000 attendees then.

lornajane:
I commented on the related ticket that I have some reservations about this feature.  I think we need to discuss and probably reframe this change.

PR #268, opened 2015-10-02
JOINDIN-565 #close Star/unstar a talk from the the details page.

petemcfarlane:


akrabat:
Loading the page doesn't pickup if the talk is already starred, in which case it should be filled int.

Also, it'd be helpful if you could rebase the branch against master.

petemcfarlane:
Thanks Rob, my bad, I've fixed the page picking up if the task is starred or not. And rebased against master.
Cheers, Pete

akrabat:
Thanks! It's nice to have fixed that inconsistency!

PR #271, opened 2015-10-02
Joindin 532 apostrophe in event

mattparker:
I don't really like this... but here it is.  Special characters are encoded and saved into the database; I can't see how you could change this without breaking too many other things.

So this decodes in the EventEntity and adds some escaping in the twig templates.


akrabat:
ok to test

akrabat:
I need to look at this one in more detail.

PR #281, opened 2015-10-07
Joindin 624

cal-tec:
Alter template to make Social Events span across all tracks in the giid view in the same way Keynotes do.
Also alter talk.type comparisons to use lower case strings in the template as the api code seems to show "Social event" as a valid talk type yet the api is producing "Social Event" as an talk type for events

http://api.joind.in/v2.1/talks/15553


akrabat:
ok to test

PR #285, opened 2015-10-09
Adding social meta for event image

abdala:
By sharing links on social network sometimes the system picks up the wrong image. This tags forces the image that will appear in the post.

But this is still not good enough. The image size is very low and I think we have to add on other pages as well. I don't understand why we don't have (removed) the speaker's image in the talk details page.

We can add another metas too. ex:

```html
<meta name="twitter:site" content="@user">
<meta name="twitter:creator" content="@user">
<meta name="twitter:url" content="http://url.com">

<meta property="og:url" content="http://url.com">
<meta property="og:title" content="title">
<meta property="og:description" content="desc">

<meta name="description" content="desc">
```

### DIFF
--- a/app/templates/Event/details.html.twig
+++ b/app/templates/Event/details.html.twig
@@ -61,6 +61,7 @@
 {% endblock %}
 
 {% block extra_meta %}
+    {{ parent() }}
     <meta property="og:image" content="{{ event.getIcon|img_path("event_icons") }}">
     <meta name="twitter:image" content="{{ event.getIcon|img_path("event_icons") }}" data-page-subject="true">
     <meta name="twitter:card" content="photo" data-page-subject="true">
--- a/app/templates/Event/details.html.twig
+++ b/app/templates/Event/details.html.twig
@@ -61,7 +61,9 @@
 {% endblock %}
 
 {% block extra_meta %}
-    {{ parent() }}
+    <meta property="og:image" content="{{ event.getIcon|img_path("event_icons") }}">
+    <meta name="twitter:image" content="{{ event.getIcon|img_path("event_icons") }}" data-page-subject="true">
+    <meta name="twitter:card" content="photo" data-page-subject="true">
 {% endblock %}
 
 {% block extra_styles %}


app/templates/Event/details.html.twig
```diff
@@ -61,7 +61,9 @@
 {% endblock %}
 
 {% block extra_meta %}
-    {{ parent() }}
```
akrabat: Why have you removed this line?

abdala: I thought it was unnecessary. We extends this view in some place?

akrabat: It's the other way around. We inherit from layout in this view, so want to pick up any extra_meta in there. I think it's empty at the moment, but we can't assume that it will stay empty in the future.

abdala: Yeah! I'll add again.

akrabat:
That looks better. Thanks!

akrabat:
ok to test

akrabat:
The URLs need to be fully qualified, not relative.

abdala:
It is. No?
screenshot 2015-10-30 11 00 02

akrabat:
Ah. I was looking at the case when there's no image provided:

```html
    <meta property="og:image" content="/img/event_icons/none.png">
    <meta name="twitter:image" content="/img/event_icons/none.png" data-page-subject="true">
    <meta name="twitter:card" content="photo" data-page-subject="true">
```

abdala:
Huuumm. It's because that if: https://github.com/joindin/joindin-web2/blob/master/app/src/View/Filters.php#L28

What's the best strategy in this case? Another filter? full_img_path()?

akrabat:
That's a good question. I'll have to think about it!

akrabat:
My initial thought is that https://github.com/joindin/joindin-web2/blob/master/app/src/View/Filters.php#L30 should return a fully qualified URL. Note though, that the domain has to be dynamically determined.

PR #288, opened 2015-10-11
JOINDIN-626 #close Enable support for web1 event urls

cal-tec:


app/src/Event/EventApi.php
```diff
@@ -115,11 +115,29 @@ public function getEvent($event_uri, $verbose = true)
...
+     * @return EventEntity
+     */
+    public function getEventById($eventId)
+    {
+        $eventId = (int)$eventId;
+        if (!$eventId) {
+            return;
+        }
+
+        $userUrl = $this->baseApiUrl . '/v2.1/events/' . $eventId;
```
akrabat: I'd rather this was called `$eventUrl`…

cal-tec: So would I.... I will fix it now

PR #291, opened 2015-10-15
New: Added compatibility route for /talk/view/<id> URL from web1

mheap:
From Jira:

> web1's long form talk urls of the format {{talk/view/15425}} need to work on web2

> This should be implemented with a new route that reuses the {{TalkController}}'s {{quickById}} action


akrabat:
add to whitelist

PR #292, opened 2015-10-24
Fixed Event Pending Description so the description body isn't wrapping around the buttons

JarrodRizor:
I changed the margin-left from 120px to 150px in the joindin.css file.


lornajane:
ok to test

PR #304, opened 2015-12-13
Add ability to upload an event icon

akrabat:
Add icon upload to the edit event form.

Also enable configuration of the base URL from which to serve images, so that we can see that it's working in the development environment (if you change you config.php as per the dist file).

This is the web2 part of [JOINDIN-530](https://joindin.jira.com/browse/JOINDIN-530).

Note: only merge this if https://github.com/joindin/joindin-api/pull/289 is accepted! Also, this PR should be updated to remove the `image_base_url` config setting if https://github.com/joindin/joindin-api/pull/292 is accepted.

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -19,6 +19,7 @@ $config = array(
         'twig' => array(
             // set to the folder name to cache into, or false to disable
             'cache' => false
-        )
+        ),
+        'image_base_url' => 'http://dev.joind.in/inc',
     ),
 );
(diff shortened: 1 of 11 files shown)


config/config.php.dist
```diff
@@ -19,6 +19,7 @@ $config = array(
         'twig' => array(
             // set to the folder name to cache into, or false to disable
             'cache' => false
-        )
+        ),
+        'image_base_url' => 'http://dev.joind.in/inc',
```
magicmonkey: Will it actually be getting the full URI from the API response?

akrabat: Separate PR for that: https://github.com/joindin/joindin-api/pull/292

magicmonkey: Ah, good good, ta.

lornajane:
Replacing this PR with https://github.com/joindin/joindin-web2/pull/336 which extends this one.

PR #317, opened 2015-12-21
Send a 410 for feed URLs

akrabat:
We don't support /feed URLs from the legacy site any more, so inform clients that they are gone.


web/.htaccess
```diff
@@ -10,6 +10,10 @@ RewriteEngine On
 
 RewriteRule ^inc/img/rating-(\d)\.gif$ img/avg-rating-$1.png [L]
 
+# Send 410 for /feed*
+RewriteEngine on
```
magicmonkey: This line doesn't need to be here, especially not in the .htaccess file.

PR #319, opened 2015-12-30
Enable listing of previous events

akrabat:
Use the API's "all" filter for the events list so that we can go backwards as well as forwards

If the page query parameter is not set, then do not pass a start number to the API's so that the event's "all" filter will return the first page of events where at least one event is in the future. That is, we get a list of events positioned at "now".

In this case, we have to work out the current page number ourselves, which we do by looking at the this_page meta field which contains the correct start parameter.

Depends on https://github.com/joindin/joindin-api/pull/295

Fixes [JOINDIN-499](https://joindin.jira.com/browse/JOINDIN-499)

### DIFF
--- a/app/src/Event/EventController.php
+++ b/app/src/Event/EventController.php
@@ -53,31 +53,25 @@ class EventController extends BaseController
 
     public function index()
     {
-        $page = $this->application->request()->get('page');
-        if ($page === null) {
-            // if page is not set, then do not set $start so that the API will return
-            // the first page of events where at least one event is in the future
-            $page = 1;
-            $start = null;
-        } else {
-            $page = ((int)$page === 0) ? 1 : $page;
-            $start = ($page -1) * $this->itemsPerPage;
-        }
+        $start = $this->application->request()->get('start');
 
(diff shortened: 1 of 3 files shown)


akrabat:
Updated to expose the `start` API parameter into the URL as it is no longer on the page boundary as per https://github.com/joindin/joindin-api/pull/295#issuecomment-172347274.

This means that when you view the events list page, the very first event is now the next event that will take place. However, if you paginate all the way back to the very start, we'll lose track as we can't have a negative start number and if you then go forwards again, when you get to "now", the next event that will take place won't be the first item on the page.

lornajane:
Should I review this as is? I know we discussed page number changes but I'm not sure whether the outcome was that this version is what we want

akrabat:
Pretty sure that I need to look at this one again to see if we can keep track of where the middle is as per https://github.com/joindin/joindin-api/pull/295#issuecomment-192949450

akrabat:
I've updated to use the `?page=1` query string format again. Does this work better for you than using `?start=46`?

PR #320, opened 2015-12-30
Highlight current submenu item on event page

akrabat:
Expose the current route to Twig and then use this to set the active class for the submenu navigation on the event page. Override the default active style to simply bold the current submenu item.

<s>Note that this will need changing if #321 is merged.</s>


akrabat:
Rebased and updated now that #321 is merged.

PR #322, opened 2016-01-01
Display reported comments

lornajane:
This just adds a "reported comments" link for event and site admins to
the event navigation, and shows the comments (using the existing comment
template) with information about who reported them.


akrabat:
Note: Once merged, we need to link to this new page in the email that goes to the administrators when a comment is reported.

PR #325, opened 2016-01-05
Added section to embed the profile

AurelioDeRosa:


AurelioDeRosa:
Before merging this PR, a couple of notes:

- The PR contains a non-existent reference to a CDN that hosts the library. I think that in combination to give the users the ability to copy and paste the code from the website, we should also provide such CDN.
- We should merge and publish this PR only when I'm ready to release version 0.6.0 of the library which will include the new `data-url` attribute.

AurelioDeRosa:
We also have to find a solution to propose to our users the possibility to choose between the themes supported by the library.

akrabat:
add to whitelist

akrabat:
Would it be useful to add some JS so that if I click on the code, it changes to a textarea with the code highlighted so that I can just press ctrl+c ? (and if I click anywhere else, it goes back to just a <code> snippet)

AurelioDeRosa:
@akrabat Done!

AurelioDeRosa:
Hi @akrabat. Any news on this PR?

akrabat:
@AurelioDeRosa I've been really busy and haven't had any time to get to this. In theory, I may have some time this weekend though.

akrabat:
I've managed bring this down locally and have looked at it now.

Can we put it into a popup type thing that is accessible via a link in the side bar please? The link in the sidebar should be in its own section.

lornajane:
@AurelioDeRosa should we keep this pull request open?  Will you be able to respond to @akrabat's requests?

AurelioDeRosa:
Hi @akrabat and @lornajane. I totally missed your replies, that's why I didn't come back to you.

Personally, I think that putting the widget in a popup will reduce the conversions, however I'm not totally against it. I can submit a PR to address the change tomorrow. Were you thinking something like [Speaker Deck](https://speakerdeck.com/aurelioderosa/improve-the-style-and-quality-of-your-project-automatically-topconf-tallinn-2015) does (click on the Embed link)?

AurelioDeRosa:
@akrabat and @lornajane Let me know if the type of popup you were looking for is similar to the one Speaker deck has so that I can apply the change you asked. Thank you

akrabat:
@AurelioDeRosa Yes - something like that would be great. A message about pressing cmd/ctrl+C to copy would be nice and I'd like it to disappear when escape is pressed.

AurelioDeRosa:
PR updated. Please, have a look and let me know your thoughts.

akrabat:
That's nice. Thanks.

One thing: can we auto-select the text, so I can just press control+c?

AurelioDeRosa:
Hi @akrabat.

The text is autoselected as soon as you click it. Changing the focus automatically might confuse some users. I hope this makes sense.

akrabat:
Hmm.. that doesn't work for me. Also, as it's not a texture, I can't work out how to select the code without using a mouse.

AurelioDeRosa:
Hi @akrabat. You were absolutely right, I forgot to update a class name. I've update the PR and this time it should be ready to go. Please note that  I've also rebased so that it can be merged easily. Let me know your thoughts.

akrabat:
Thanks!

AurelioDeRosa:
Hi @akrabat.

As I mentioned in one of my first messages, the PR contains a non-existent reference to a CDN that hosts the library. Do we have a real CDN we can use to embed the library? Otherwise the code won't work.

akrabat:
aargh! I missed that…

PR #327, opened 2016-01-08
Changes max number of displayed CfPs to 4

heiglandreas:
This commit changes the number of CfPs displayed on the main page from 10 to 4. That way the site doesn't look that cluttered anymore.


lornajane:
A small change but it makes a big difference!  Thanks

PR #328, opened 2016-01-08
Fixes broken display of CfP-Dates

heiglandreas:
Currently in the list of upcomming CfPs there is no information given to when the CfP starts and/or ends. This is due to wrongly named methods for CfP-Date-retrieval. I've fixed these function names and now the CfP-Info is displayed again.


lornajane:
Much better, thanks :)

PR #329, opened 2016-01-08
Adds a link to callingallpapers.com

heiglandreas:
This commit adds a link to callingallpapers.com as alternative to the link to the CfP-Page on joind.in


lornajane:
These are technically independent projects, and plenty of other cfp tools also use joind.in, so I'd prefer not to add the link.

heiglandreas:
A pity but I can absolutely understand that :wink:

PR #331, opened 2016-02-02
Moderate a comment

akrabat:
Add Accept and Reject buttons to each comment on the reported comments page.

Depends on [API PR 308](https://github.com/joindin/joindin-api/pull/308)

### DIFF
--- a/app/src/Event/EventApi.php
+++ b/app/src/Event/EventApi.php
@@ -425,4 +425,27 @@ class EventApi extends BaseApi
 
         return $reports;
     }
+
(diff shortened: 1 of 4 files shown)


lornajane:
When I try accepting or rejecting comments as an event host: I can see the list of comments and the accept/reject buttons but clicking them gives me the error message "We're sorry, but you are not authorised to do that."  As a site admin, I can accept and reject the reports fine.

akrabat:
Sorry! Fixed.

PR #333, opened 2016-02-14
Fixes number of CfPs on Event-subpage

heiglandreas:
Currently on the main entry page only 4 CfPs are shown. Until now the /event-page showed 10 CfPs as that one slipped my eyes when I altered showing only 4 CfPs. 

So this is the fix to that.


akrabat:
Thanks.

PR #335, opened 2016-02-28
Add admin nav-bar to event page

akrabat:
With both Edit and Reported comments already available and the need to add addtional admin-only menu items, it makes sense to move them to their own nav-bar under the attendee-facing nav-bar.

I've also added the nav-bar to the edit event page as it was missing. This meant separating the nav-bar rendering from the event header template, so I could reuse it.

Finally, for consistency, I've also moved the nav-bar on the Events listing page to below the title too and made it so that there's always a border at the bottom of a nav-bar, but never a situation where we have two borders with no text between them.

### DIFF
--- a/app/templates/Event/_common/event_header.html.twig
+++ b/app/templates/Event/_common/event_header.html.twig
@@ -17,33 +17,5 @@
             </h1>
         </div>
     </div>
-    <nav class="page-nav">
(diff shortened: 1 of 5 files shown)


lornajane:
Closing as we have the admin bar coming in with 338 in any case

PR #337, opened 2016-03-25
Fix style of date fields on event form

akrabat:
The date fields on the event form need the `form-control` class in order to match the style of all the other elements on the form.


lornajane:
Such a nice touch, thanks :)

PR #338, opened 2016-03-25
Add talk to an event

akrabat:
Add the ability to add a talk to an event. 

This PR also implements the yellow admin bar for the event section as we now have three items for it and will have more, so this seemed a good time to to do it.

The ability to add speakers when adding the talk is also implemented. This works the same way as web1 and depends on [API PR #320](https://github.com/joindin/joindin-api/pull/320).

In order for the talk to be added to the selected track, [API PR #328](https://github.com/joindin/joindin-api/pull/328) is required.

### DIFF
--- a/app/src/Event/EventController.php
+++ b/app/src/Event/EventController.php
@@ -843,6 +843,10 @@ class EventController extends BaseController
         if (!$event) {
             return Slim::getInstance()->notFound();
         }
+        if (!$event->getCanEdit()) {
+            $this->application->flash('error', "You do not have permission to do this.");
+            $this->redirectToDetailPage($event->getUrlFriendlyName());
+        }
 
         $languageApi = $this->getLanguageApi();
         $languages = $languageApi->getLanguagesChoiceList();
--- a/app/src/Event/EventController.php
+++ b/app/src/Event/EventController.php
@@ -612,15 +612,15 @@ class EventController extends BaseController
      */
     private function addEventUsingForm(Form $form)
     {
-        $eventapi = $this->geteventapi();
-        $values = $form->getdata();
+        $eventApi = $this->getEventApi();
+        $values = $form->getData();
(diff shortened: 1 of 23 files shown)


akrabat:
hmm... I've just realised that there's no way to add a talk to a track when creating it. 

@lornajane: How do you want to attach a talk to a track?

lornajane:
Since we can't create tracks, this is probably out of scope this time
around?

lornajane:
This PR is now showing conflicts, can you have a quick look?

akrabat:
Rebased now that #331 has been merged.

app/src/Event/EventController.php
```diff
@@ -606,15 +612,15 @@ public function redirectFromId($eventId, $extra = false)
      */
     private function addEventUsingForm(Form $form)
     {
-        $eventApi = $this->getEventApi();
-        $values = $form->getData();
+        $eventapi = $this->geteventapi();
```
Jean85: What are these changes? Unwanted "Toggle caps"?

Jean85:
LGTM.

I've tested this locally and it seems to work. 
The only quirk that I found was the timepicker helper in the form: if I click on the field, the picker comes up with a prefilled value (e.g. 9:00AM) but the field itself stays empty until I change it; so, if the prefilled value was ok for me, I have to move the time up\down and then back to the prefilled value to make it work.

app/src/Event/EventController.php
akrabat: Good catch. Fixed.

liam-wiltshire:
I'm officially on holiday now (whooo!), so I'll try to check and merge some of these larger PRs this week.

akrabat:
Rebased and ran `composer dumpautoload` in order to pick up and update `autoload_static.php`.

PR #339, opened 2016-03-26
Track administration

akrabat:
Ability to add, edit and delete tracks from an event.

This is implemented as a page containing a form with a collection of track form-types in it, so we can implement the functionality as a single page.

Requires PR [#338 (Add talk to an event)](https://github.com/joindin/joindin-web2/pull/338) as that one has the new admin nav bar in it.


akrabat:
When #338 has been merged, I'll rebase this one...

PR #340, opened 2016-03-27
Edit talk

akrabat:
Add the ability to edit a talk from the talk's detail. 

I haven't put a menu bar as there's only one item and it looked silly, so I've put an "Edit talk" button under the description. Thoughts welcome on if this works.

This PR is built on top of #338  (Add talk to an event).


akrabat:
Rebased now that #338 has been merged.

PR #342, opened 2016-04-14
Added 15px bottom padding to row class for better vertical spacing

kennyray:
I added 15px to the bottom of the row class. I think 15px gives it nice vertical spacing. I tried 5,10, and 15, and I think 15 looks best at all widths.


akrabat:
okay to test

akrabat:
Thanks :)

PR #343, opened 2016-07-18
Require security patched version of guzzle

AndrewCarterUK:
Regarding: https://httpoxy.org/


akrabat:
ok to test

composer.json
```diff
@@ -24,6 +24,6 @@
         "symfony/security-csrf": "~2.5",
         "symfony/twig-bridge":   "~2.5",
         "org_heigl/daterange": "~1.1",
-        "guzzlehttp/guzzle": "^6.1"
+        "guzzlehttp/guzzle": "~6.2.1"
```
akrabat: Why is this change here?
Unless, I've misunderstood, ~6.2.1 means that we won't pick up 6.3.0.

AndrewCarterUK: I released that the addition of the guzzle dependency was recent (as I had to update my fork to get it). As all the other dependencies used the '~' operator rather than '^', I assumed that tighter version constraints was a project preference.

What version constraint should I change it to?

akrabat: I think that the `~` in the other options indicates when they were added and for a `major.minor` constraint it doesn't make any practical difference if you use `~` or `^`. However if you use a `major.minor.patch` constraint, then it does make a difference and `^` is preferred.

AndrewCarterUK:
I've updated the branch

akrabat:
Thanks for this one too :)

PR #344, opened 2016-08-19
Adding icon submission guidelines (solves JOINDIN-695)

Jean85:
As in title.

I've added a guideline in the sidebar as suggested in the issue, and also a short help message below the upload field in the form, so it's immediately visible, even on mobile.


akrabat:
add to whitelist

PR #347, opened 2016-08-21
Keep comment if no rating validation fails - solves JOINDIN-684

liam-wiltshire:
Solves JOINDIN-684 by ensuring that the comment is re-populated into the form if no rating was provided, and vice-versa (maintains the rating if no comment was provided).

Also ensures that the correct anchor is returned in all error instances (Spam, Duplicated, etc.)


akrabat:
add to whitelist

akrabat:
👍

PR #348, opened 2016-08-23
Adding a default apple-touch-icon (solves JOINDIN-672)

Jean85:
Jira: https://joindin.jira.com/browse/JOINDIN-672

Following Apple's docs:

> If no icons are specified using a link element, the website root directory is searched for icons with the apple-touch-icon... prefix. For example, if the appropriate icon size for the device is 60 x 60, the system searches for filenames in the following order:
> 
>  1. apple-touch-icon-76x76.png
>  2. apple-touch-icon.png

I've added a default icon, with nop modification of the template needed; it's created from a `iphone.jpg` file in the legacy repo.


heiglandreas:
Thanks for the contribution!

According to [Apples web-application documentation](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4) and [Apples icon-size documentation](https://developer.apple.com/ios/human-interface-guidelines/graphics/app-icon/) the icon should be either 152 or 120 or 76 pixel wide and high. And you can also provide different sized icons.

Would you mind adapting the PR accordingly?

Jean85:
@heiglandreas done!

heiglandreas:
Shouldn't there also be the referenced ```<link … />``` entry in https://github.com/joindin/joindin-web2/blob/master/app/templates/layout.html.twig? That would ease the load on the server as the client would know what icons are available and wouldn't need to request icons until a suitable one is found. 

In my eyes there should be something like this in the layout-file:

    <link rel="apple-touch-icon" href="apple-touch-icon.png">
    <link rel="apple-touch-icon" sizes="76x76" href="apple-touch-icon-76x76.png">
    <link rel="apple-touch-icon" sizes="120x120" href="apple-touch-icon-120x120.png">
    <link rel="apple-touch-icon" sizes="152x152" href="apple-touch-icon-152x152.png">

The files could also be stored in ```img``` and named according to the current scheme: ```app-iphone-[60|76|120|152].png``` (as an example). That would require the new image-files to b e relocated though :wink:

Jean85:
Umh yeah you're right, I didn't think about the server load.
Since I'm adding the `<link>` tags, I'll move the images too, then!

Jean85:
Thanks @heiglandreas for the suggestions.

Done. I moved the files, but I preferred not to change the file name scheme since in the `img` folder we already have `apple-*` files, and `app-iphone-*` too, but it's a screenshot of the app (maybe for the App Store?).

I've also added new icon sizes, since Apple's documentations refers the for newer models.

akrabat:
FWiW, the screenshot image is for https://joind.in/apps

heiglandreas:
Looks great! Thanks!

From my POV this could be merged.

liam-wiltshire:
We could do with one for Android as well (uses <link rel="icon" /> afaik), but will merge this as is, and add an issue to Jira (potentially a hackathon item)

PR #349, opened 2016-08-23
Delete User Frontend and Controllers

liam-wiltshire:
Re-built the branch without all the messed up commits :-)


liam-wiltshire:
ok to test

PR #350, opened 2016-08-24
Add who's attending block to event details page - satisfies JOINDIN-621

liam-wiltshire:
Creates a basic sidebar block to show event attendees - shows 6 to start, with the rest loaded via AJAX on request (otherwise there could potentially be a pretty substantial performance hit for big events!)

The styling is fairly basic - it works and looks 'fine' I guess, but no doubt someone could pick it up and make it look better once it's in.

### DIFF
--- a/app/src/Event/EventApi.php
+++ b/app/src/Event/EventApi.php
@@ -5,6 +5,7 @@ use Application\BaseApi;
 use Talk\TalkCommentEntity;
 use Talk\TalkCommentReportEntity;
 use User\UserApi;
+use User\UserEntity;
 
 class EventApi extends BaseApi
 {
@@ -206,6 +207,33 @@ class EventApi extends BaseApi
         }
 
         throw new \Exception("Failed to unmark you as attending: " . $result);
+    }
+
+
+    /**
+     * Get attendees for given event
+     * @param $attendees_uri
+     * @param bool $verbose
+     * @return UserEntity[]
+     */
+    public function getAttendees($attendees_uri, $limit = 0, $verbose = false)
+    {
+
+        $attendees_uri .= "?resultsperpage={$limit}";
+        if ($verbose) {
+            $attendees_uri = $attendees_uri . '&verbose=yes';
+        }
+
+
+        $attendees = (array)json_decode($this->apiGet($attendees_uri));
+
+        $attendeeData = array();
+
+        foreach ($attendees['users'] as $attendee) {
+            $attendeeData[] = new UserEntity($attendee);
+        }
+
+        return $attendeeData;
     }
 
     /**
--- a/app/src/Event/EventApi.php
+++ b/app/src/Event/EventApi.php
@@ -213,6 +213,7 @@ class EventApi extends BaseApi
     /**
      * Get attendees for given event
      * @param $attendees_uri
+     * @param $limit
      * @param bool $verbose
      * @return UserEntity[]
      */
--- a/app/src/Event/EventController.php
+++ b/app/src/Event/EventController.php
@@ -192,7 +192,7 @@ class EventController extends BaseController
             . $this->application->urlFor('event-quicklink', array('stub' => $event->getStub()));
 
 
-        $attendees  = $eventApi->getAttendees($event->getAttendeesUri(),6);
+        $attendees  = $eventApi->getAttendees($event->getAttendeesUri(), 6);
 
         $this->render(
             'Event/details.html.twig',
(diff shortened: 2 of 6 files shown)


app/src/Event/EventApi.php
```diff
@@ -208,6 +209,34 @@ public function unattend(EventEntity $event)
...
+     * @return UserEntity[]
+     */
+    public function getAttendees($attendees_uri, $limit = 0, $verbose = false)
+    {
+
+        $attendees_uri .= "?resultsperpage={$limit}";
+        if ($verbose) {
+            $attendees_uri = $attendees_uri . '&verbose=yes';
+        }
+
```
Jean85: A lot of empty newlines here and later... Can you clean those up?

app/src/Event/EventEntity.php
```diff
@@ -541,6 +541,11 @@ public function getReportedTalkCommentsUri()
         }
     }
 
+    public function getAttendeesUri(){
```
Jean85: Please respect PSR-2... You can use the phing targets to lint your code!

app/templates/Event/_common/event_attendees.html.twig
```diff
@@ -0,0 +1,17 @@
+{% if (event.getAttendeeCount  > 0) %}
+<h3>Who's attending?</h3>
+    <div class="row">
+        {% for attendee in attendees %}
+            <div class="col col-sm-6 col-xs-4">
```
Jean85: I would suggest adding a `text-center' class to this `<div>`, it will look a lot nicer!

akrabat:
add to whitelist

app/src/Event/EventEntity.php
akrabat: The Jenkins build will pick up PSR-2 issues, so no need to comment on them.

web/css/joindin.css
```diff
@@ -624,6 +624,19 @@ a.talk.star {
...
+.event-attendees img {
+    max-width:100%
+}
+
+.event-attendees .row .col {
+    text-align:center;
+}
+
+.event-attendees .row .col .attendeeName {
+    height:40px;
```
Jean85: This seems an overkill... I tried replacing that with a
```
padding-top: 2px;
padding-bottom: 10px;
```
and its seems better to me

app/src/Event/EventController.php
```diff
@@ -182,18 +183,44 @@ public function details($friendly_name)
...
+    public function attendees($friendly_name){
+        $eventApi = $this->getEventApi();
+        $event    = $eventApi->getByFriendlyUrl($friendly_name);
+
+        if (! $event) {
+            return Slim::getInstance()->notFound();
+        }
+
+
+        $attendees  = $eventApi->getAttendees($event->getAttendeesUri());
```
Jean85: This will include the logged user if attending. Is that correct/intended?

app/src/Event/EventApi.php
akrabat: @Jean85 Don't worry too much about whitespace. If it doesn't break PSR-2, it's probably okay. The main thing to focus on is that the PR works as intended, especially for use-cases that maybe the submitter may not have thought to check.

web/css/joindin.css
liam-wiltshire: @Jean85 - this has a fixed height to get round issues with names that wrap onto two lines - keeping the name div consistent to support longer names, without having other issues with broken floats etc, so I think it does need to be height:40px

Jean85: Ah you're right! I forgot to check that.
I've played with that a little more, I was just bothered by the fact that there's no spacing between the name and the avatar. Can I suggest this approach then?
```
    height: 42px;
    padding-top: 2px;
```

Jean85:
Screenshots of the results: http://imgur.com/a/aDota

Jean85:
This seems ready to be merged, isn't it?

akrabat:
Each attendee seems a little large to me & as such I'm not sure this will scale to 100 attendees. Can we do say 4 per row?

liam-wiltshire:
I think 4 per row is too many, as it ends up chopping off bits of names etc

Having a play, I think 3 across fits, so I'll update it to that for now.

If in future someone with better CSS than I can come up with a more elegant solution for the 'long names' problem, then that'd be good :-)

app/templates/Event/_common/event_attendees.html.twig
```diff
@@ -0,0 +1,17 @@
+{% if (event.getAttendeeCount  > 0) %}
+<h3>Who's attending?</h3>
+    <div class="row">
+        {% for attendee in attendees %}
+            <div class="col col-sm-4 col-xs-3">
```
heiglandreas: Wouldn't it be possible to set this to ```col-lg-3 col-md-4 col-sm-6 col-xs-12```? Then it should be somewhat responsive and we might solve the question @akrabat raised regarding scalability for 100 users :wink: 

Alternative would be a list with the ```li``` styled as inline-block.

I'D also here prefer a styling using ```<ul>``` as it's a list here, but that's absolutely minor.

PR #351, opened 2016-09-07
Ensure user is logged in for canEditTalk flag

akrabat:
If the user is not logged in, don't check if the talk can be edited.

Fixes 500 error that's currently one live!


Jean85:
LGTM :+1: 
I was writing the same fix rn :laughing:

PR #354, opened 2016-09-30
Hides rating bar in add comment box if you have already commented.

railto:
Hides rating bar in add comment box if you have already commented.  Sends default rating value if adding an additional comment so the API accepts the comment.

### DIFF
--- a/app/src/Talk/TalkCommentEntity.php
+++ b/app/src/Talk/TalkCommentEntity.php
@@ -123,4 +123,13 @@ class TalkCommentEntity
 
         return $this->data->user_uri;
     }
+
+    public function canRateTalk($user_uri)
+    {
+        if ($this->data->user_uri == $user_uri) {
+            return false;
+        }
+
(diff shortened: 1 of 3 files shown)


liam-wiltshire:
@railto,

Works great, any chance you could extend it to not show the rating bar if the person commenting is the speaker please? Know it wasn't on the ticket, but would be a nice thing to add :-)

Having had a look at the code, it should just be a case of adding `$talk->isSpeaker($_SESSION['user']->getUri())` into your check I think?

Cheers!

railto:
I've looked into adding that and my concern is that the overall rating would be affected.  The API does not appear to filter out comments/ratings from the speaker (if talk is claimed that is).  Would need to update the API before adding this in.

liam-wiltshire:
@railto - this is handled in the SQL function `get_talk_rating` - https://github.com/joindin/joindin-api/blob/master/db/patch66.sql

PR #362, opened 2016-10-01
Revert "Listing 5 hot events on the about and apps page."

desaijay:


liam-wiltshire:
Hi @desaijay 

I don't really know much about this PR - is this based on an issue either in GitHub or in JIRA?

liam-wiltshire:
This just appears to be necromancy of an old commit that doesn't achieve anything - if I've missed anything, please do open a new PR, however with no reply to my previous message I'm going to close for now.

PR #367, opened 2016-10-07
CSS Organisation

marktcotton:
Started breaking CSS into sections with a view to add a frontend process and split into smaller files that are compiled. 

Ideally this will utilise the already included Grunt to make sure CSS / JS is compiled, minified and linted.


liam-wiltshire:
@marktcotton this seems to remove the fixed with on the container over 1200px - was this intentional?

`@media (min-width: 1200px)
.container-fluid {
    width: 1170px;
}`

doesn't seem  to be in the refactored version?

clairs:
Just found this PR. Just a thought, should we convert to SASS instead? Would make it far easier to split the CSS into sections/separate files.

liam-wiltshire:
@clairs  I'm more than happy to move everything to SASS - as long as someone who knows far more about SASS than I can maintain it :-)

railto:
@marktcotton Sorry for the massive delay in getting back to you on this.  I'm going to close this in favour of #408 (hopefully we can get that one sorted as well)

PR #369, opened 2016-10-14
Composer improvements

ericpoe:
This started out as a search to figure out which version of PHP I should write against as a new joind.in dev. I first looked in the `CONTRIBUTING.md` file and didn't find anything. I then looked at the `composer.json` file and didn't see anything. However,`"preferred-install": "dist"` being a top-level setting in `composer.json` appeared to be incorrect, so I looked into where it needed to be: it needs to be in a config.

This PR attempts to fix both issues.

As a bonus, I guess, Composer 1.2.1 (latest version at this time) wanted to make some changes to the logic in Classloader. Those changes are part of this PR as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,3 +49,7 @@ An example would be:
 ```
 JOINDIN-445 #close Fixed.
 ```
+
+## PHP Version
+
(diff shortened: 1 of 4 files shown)


liam-wiltshire:
ok to test

PR #371, opened 2016-10-20
Fixes possibly endles while-loop in comment-field

heiglandreas:
In some edge-cases it was possible that the while-loop that adapted the height of the comment field to the contents height went berserk and extended the height indefinitely. This commit changes that loop to one single function call so it will not be able to run indefinitely and leave the browser unresponsive.

Thanks to Elizabeth Smith for finding the issue and testing the fix.

Fixes JOINDIN-713

### DIFF
--- a/app/templates/Talk/index.html.twig
+++ b/app/templates/Talk/index.html.twig
@@ -13,9 +13,8 @@
         });
         $('#rating_widget').bind('rated', function (event, value) { $("#rating_value").html($("#rating option[value='"+value+"']").text()) });
         $("#comment").keyup(function(e) {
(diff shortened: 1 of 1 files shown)


liam-wiltshire:
@heiglandreas - looks good, to you have a specific way to reproduce the edge case? Would like to be able to give it a quick test :-)

heiglandreas:
I'm pretty sorry, but I couldn't even test it my own machine. I had a slight idea what it might have been and send Elizabeth the changed javascript, she tested it in her machine and it worked. And as I couldn't find issues with the altered code on my browsers I was happy :)

So no, there's no way to reproduce it apart from Elizabeths laptop (which has a touch-screen, so that might be the issue...)

PR #372, opened 2016-10-21
Fixing redirect after not-allowed action JOINDIN-701

lucianoqueiroz:
I searched across the project usages of "not-allowed" page, and just the "user-profile-edit" was missing the "redirect" parameter.


liam-wiltshire:
ok to test

PR #374, opened 2016-10-25
Use abstract classes

ericpoe:
This is a code cleanup PR.

This formally declares two classes that were being used as abstract classes as abstract. This also extracts out some more abstract classes to make it easier to identify differences between like-classes.

I found a logic bug the EventScheduler while I was running through this, so I've also included its fix in this PR. If you'd prefer that fix be in a separate PR, let me know and I will split it out. (edit: split out to #375 )

### DIFF
--- a/app/src/Application/BaseController.php
+++ b/app/src/Application/BaseController.php
@@ -18,7 +18,10 @@ abstract class BaseController
         $this->defineRoutes($app);
         $this->cfg = $this->getConfig();
 
-        $this->accessToken = isset($_SESSION['access_token']) ? $_SESSION['access_token'] : null;
+        $this->accessToken = null;
+        if (isset($_SESSION['access_token'])) {
+            $this->accessToken = $_SESSION['access_token'];
+        }
     }
 
     private function getConfig()
--- a/app/src/Application/BaseController.php
+++ b/app/src/Application/BaseController.php
@@ -7,7 +7,7 @@ use Twig_Error_Runtime;
 abstract class BaseController
 {
     /** @var Slim */
-    protected $application = null;
+    protected $application;
 
     protected $accessToken;
     protected $cfg;
--- a/app/src/Application/BaseEntity.php
+++ b/app/src/Application/BaseEntity.php
@@ -1,15 +1,17 @@
 <?php
 namespace Application;
 
+use stdClass;
+
 abstract class BaseEntity
 {
     protected $data;
 
     /**
      * BaseEntity constructor.
-     * @param \stdClass $data Model data retrieved from API
+     * @param stdClass $data Model data retrieved from API
      */
-    public function __construct($data)
+    public function __construct(stdClass $data)
     {
         $this->data = $data;
     }
--- a/app/src/Application/BaseEntity.php
+++ b/app/src/Application/BaseEntity.php
@@ -1,0 +1,16 @@
+<?php
+namespace Application;
+
+abstract class BaseEntity
+{
+    protected $data;
+
+    /**
+     * BaseEntity constructor.
+     * @param \stdClass $data Model data retrieved from API
+     */
+    public function __construct($data)
+    {
+        $this->data = $data;
+    }
+}
--- a/app/src/Application/BaseEntity.php
+++ b/app/src/Application/BaseEntity.php
@@ -5,6 +5,7 @@ use stdClass;
 
 abstract class BaseEntity
 {
+    /** @var stdClass */
     protected $data;
 
     /**
--- a/app/src/Event/EventCommentReportEntity.php
+++ b/app/src/Event/EventCommentReportEntity.php
@@ -1,24 +1,24 @@
 <?php
 namespace Event;
 
-class EventCommentReportEntity
-{
-    private $data;
-    private $comment;
+use Application\BaseCommentReportingEntity;
 
+class EventCommentReportEntity extends BaseCommentReportingEntity
+{
     /**
      * Create new EventCommentReportEntity
      *
-     * @param Object $data Model data retrieved from API
+     * @param \stdClass $data Model data retrieved from API
      */
-    public function __construct($data)
+    public function __construct(\stdClass $data)
     {
+        parent::__construct($data);
+
         // should contain a comment
         if ($data->comment) {
             $this->comment = new EventCommentEntity($data->comment);
             unset($data->comment);
         }
-        $this->data = $data;
     }
 
     public function getComment()
--- a/app/src/Event/EventCommentReportEntity.php
+++ b/app/src/Event/EventCommentReportEntity.php
@@ -2,15 +2,16 @@
 namespace Event;
 
 use Application\BaseCommentReportingEntity;
+use stdClass;
 
 class EventCommentReportEntity extends BaseCommentReportingEntity
 {
     /**
      * Create new EventCommentReportEntity
      *
-     * @param \stdClass $data Model data retrieved from API
+     * @param stdClass $data Model data retrieved from API
      */
-    public function __construct(\stdClass $data)
+    public function __construct(stdClass $data)
     {
         parent::__construct($data);
 
--- a/app/src/Event/EventController.php
+++ b/app/src/Event/EventController.php
@@ -30,7 +30,7 @@ class EventController extends BaseController
         $this->pendingItemsPerPage = 30;
     }
 
-    protected function defineRoutes(\Slim\Slim $app)
+    protected function defineRoutes(Slim $app)
     {
         // named routes first; should an event pick the same name then at least our actions take precedence
         $app->get('/event', array($this, 'index'))->name("events-index");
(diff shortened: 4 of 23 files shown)


liam-wiltshire:
ok to test

ericpoe:
On my drive to work this morning, I was thinking about the removal of constructors that only call to the `parent::__constructor($params...)` on the derived classes (ex. f4923b0f88d).

While technically correct that derived classes will infer the constructor from the parent, I am concerned that this might be a hinderance for other contributors who will need to crawl up the class tree to find out which parameters the class requires during instantiation.

I would like for this to match the opinion of the project collaborators.

If you would like all classes to use a constructor, then I'll update this tonight.

Thank you.

app/src/Event/EventSchedulerDay.php
```diff
@@ -1,7 +1,7 @@
 <?php
 namespace Event;
 
-use DateTime;
+use \DateTime;
```
heiglandreas: Never prefix a namespace with a backslash in ```use```

heiglandreas:
Thanks for this awesome and impressive PR.

But (you knew there'd be one): For me personally there are too many different things in this PR. There are bugfixes, there is refactoring and there are codestyle-fixes all mixed into this. That makes it very hard to digest and to see wha addresses what.

I would prefer this to be split into multiple PRs. One handling the introduction of the base-classes, one for each bugfix (as they are relevant even without the refactoring! Thanks for spotting the issues!) and so on. This would make it much easier to decide on the separate things.

ericpoe:
Not a problem @heiglandreas; I knew it was a bit heavy. I'll split stuff out into separate PRs.

Jean85:
Woha great PR! Splitting would be really good, I agree.
On the matter of (emtpy) constructor, you would prefer removing them in those cases.

ericpoe:
Hi @heiglandreas and @Jean85, this PR is now just the refactor. Even though 1d3372e is not involved with abstract classes, I kept it as part of this PR since it is just pulling property initialization into the constructor -- which is something that much of this PR includes anyway.

app/src/Event/EventCommentReportEntity.php
```diff
@@ -1,24 +1,24 @@
...
+class EventCommentReportEntity extends BaseCommentReportingEntity
+{
     /**
      * Create new EventCommentReportEntity
      *
-     * @param Object $data Model data retrieved from API
+     * @param \stdClass $data Model data retrieved from API
      */
-    public function __construct($data)
+    public function __construct(\stdClass $data)
```
heiglandreas: Add a ```use stdClass;``` to the use-statements and omit the backslash here.

app/src/Event/EventController.php
```diff
@@ -20,8 +20,15 @@
 
 class EventController extends BaseController
 {
-    private $itemsPerPage = 10;
-    private $pendingItemsPerPage = 30;
+    private $itemsPerPage;
+    private $pendingItemsPerPage;
+
+    public function __construct(Slim $app)
```
heiglandreas: When we're using ```Slim``` here we should also change occurences for ```\Slim\Slim``` with ```Slim``` everywhere else in the file…

app/src/Search/SearchController.php
```diff
@@ -26,7 +25,13 @@ class SearchController extends BaseController
     /**
      * @var integer The number of search results to show per page
      */
-    protected $itemsPerPage = 10;
+    protected $itemsPerPage;
+
+    public function __construct(Slim $app)
```
heiglandreas: If we use ```Slim``` here there should also be a ```use Slim\Slim;``` in this file, even though it's already set in the BaseController-File. Makes it more readable IMHO

app/src/Talk/TalkCommentEntity.php
```diff
@@ -1,36 +1,14 @@
...
-        }
-
-        return $this->data->rating;
-    }
+use Application\BaseCommentEntity;
 
-    public function getUserDisplayName()
+class TalkCommentEntity extends BaseCommentEntity
+{
+    public function __construct(\stdClass $data)
```
heiglandreas: See above for ```use stdClass;```

app/src/Talk/TalkCommentEntity.php
```diff
@@ -1,36 +1,14 @@
...
     {
-        if (!isset($this->data->user_display_name)) {
-            return null;
-        }
-
-        return $this->data->user_display_name;
+        parent::__construct($data);
+        $this->commentUri = isset($this->data->uri) ? $this->data->uri : null;
```
heiglandreas: Also see above. But when this is present in both Comment-Classes wouldn't it be better of in the constructor of the base-class?

app/src/Talk/TalkCommentReportEntity.php
```diff
@@ -1,38 +1,23 @@
...
+class TalkCommentReportEntity extends BaseCommentReportingEntity
+{
     /**
      * Create new TalkCommentReportEntity
      *
-     * @param Object $data Model data retrieved from API
+     * @param \stdClass $data Model data retrieved from API
      */
-    public function __construct($data)
+    public function __construct(\stdClass $data)
```
heiglandreas: See above re. ```use stdClass;```

app/src/Talk/TalkDb.php
```diff
@@ -2,10 +2,15 @@
 namespace Talk;
 
 use Application\BaseDb;
+use Application\CacheService;
 
 class TalkDb extends BaseDb
 {
-    protected $keyName = 'talks';
```
heiglandreas: Shall that really be removed?

app/src/Talk/TalkCommentEntity.php
ericpoe: This checks for `$this->data->uri` and the other checks for `$data->comment_uri` (soon to be `$this->data->comment_uri`). This is a great example for why extending the base class makes for more readable code!

heiglandreas: Sorry, didn't scroll up to check the actual variable-name. You're right there. Though I'D still prefer the "long" version ;)

app/src/Application/BaseEntity.php
```diff
@@ -0,0 +1,18 @@
+<?php
+namespace Application;
+
+use stdClass;
+
+abstract class BaseEntity
+{
+    protected $data;
```
greydnls: Should there be a docblock here? `\stdClass` ?

app/src/Application/BaseController.php
```diff
@@ -18,7 +18,10 @@ public function __construct(Slim $app)
         $this->defineRoutes($app);
         $this->cfg = $this->getConfig();
 
-        $this->accessToken = isset($_SESSION['access_token']) ? $_SESSION['access_token'] : null;
```
heiglandreas: ```null``` should be the default value of the property, so it's a single ```if```-statement.

app/src/Search/SearchController.php
greydnls: :+1:

app/src/Event/EventDb.php
```diff
@@ -2,10 +2,15 @@
 namespace Event;
 
 use Application\BaseDb;
+use Application\CacheService;
 
 class EventDb extends BaseDb
 {
-    protected $keyName = 'events';
+    public function __construct(CacheService $cache)
```
ericpoe: `protected $keyName` is now in the base class. This change uses the constructor to set its value rather than setting the value outside of the constructor.

app/src/Talk/TalkCommentEntity.php
```diff
@@ -1,36 +1,17 @@
...
-    public function getUserDisplayName()
+class TalkCommentEntity extends BaseCommentEntity
+{
+    public function __construct(stdClass $data)
     {
-        if (!isset($this->data->user_display_name)) {
-            return null;
+        parent::__construct($data);
+        if (isset($this->data->uri)) {
+            $this->commentUri = $this->data->uri;
```
ericpoe: @greydnls I like it. Updated via 6abcde80b7d29e2d

ericpoe:
In rechecking this PR, I'm not seeing any requested changes that are missing. Is this waiting on some additional input or changes from me? I just want to make sure that all of my ducks are in a row. Cheers!

liam-wiltshire:
Sorry @ericpoe, with it changing quite a few things I wanted to have time to go through it properly.

Will test it today and merge.

ericpoe:
I don't want to rush this. I just wanted to make sure that this wasn't waiting for more updates from me. Please, take as much time as needed to ensure a quality merge.

liam-wiltshire:
I've written a big test matrix that I'm running through as we speak, so should be merged shortly :-)

liam-wiltshire:
Thanks for this @ericpoe - going to do a build do test this evening, but all worked locally :-)

PR #375, opened 2016-10-26
Fix error in checking for empty array

ericpoe:
This fixes a logic error where the parentheses in an if-statement were misplaced. 

This replaces a `count` being greater than zero with a check for not- `empty`, since that's really what the intent is here.


liam-wiltshire:
ok to test

PR #376, opened 2016-10-26
Fix existing PSR2 violations

ericpoe:
This is so existing violations don't trip up a future PR.

These were fixed via `phpcbf --standard=PSR2 app/`


liam-wiltshire:
ok to test

PR #377, opened 2016-10-26
Add missing property

ericpoe:
This class property existed without being explicitly declared. Was it public? private? protected? This commit resolves that question.


liam-wiltshire:
ok to test

PR #382, opened 2016-10-31
Compact login box for talk page

clairs:
Show a more compact version of the "you need to login to comment" box on the
talk page, until visitor clicks a "more" link.

[JOINDIN-691]


heiglandreas:
Now there's no mention whatsoever that one needs to log in to comment. Is that on purpose? 

Or shouldn't there be a note saying "please log in to comment. To do so use the login-panel you can find *here*" <= triggers a javascript to scroll the login field into the viewport…

clairs:
Yeah, it was done on purpose. I assume that people will figure out that they need to be registered and logged in, without having to clutter the page telling them so :)

railto:
I really think we should have some notes that the visitor needs to login to add a comment, else it may look like comments are disabled and they may then decide to not login and thus miss being able to leave a comment.

liam-wiltshire:
Ok, so I've added a line to see what it looks like - @clairs @heiglandreas @railto, if you guys can check this branch out and see how it looks, hopefully we can get a consensus!

Personally I don't think it's needed, but I don't mind either way.

railto:
Works for me, thinking maybe we should add the same to event comments page? thoughts?

clairs:
I wasn't sure how obvious it was where the login box was, as clicking the link goes to the same page. So I experimented with adding a flash of colour... what do you guys think?

heiglandreas:
Looks good for me! Clicking multiple times onto the link will not trigger the highlighting again but hey, that's something that happens during testing and not during LiveUsage!

Shouldn't that be added to the Event-Comments as well?

clairs:
Yep, I agree, let's keep it consistent. I've made the change.

PR #386, opened 2016-11-03
Add ability to unlink an approved speaker from a talk #JOINDIN-647

railto:
Work in progess, feedback appreciated

### DIFF
--- a/app/src/Talk/TalkController.php
+++ b/app/src/Talk/TalkController.php
@@ -446,6 +446,34 @@ class TalkController extends BaseController
         $this->application->redirect($url);
     }
 
+    public function unlinkSpeaker($eventSlug, $talkSlug, $username)
+    {
+        if (!isset($_SESSION['user'])) {
+            $thisUrl = $this->application->urlFor('talk', ['eventSlug' => $eventSlug, 'talkSlug' => $talkSlug]);
+            $this->application->redirect(
+                $this->application->urlFor('not-allowed') . '?redirect=' . $thisUrl
+            );
+        }
+
+        $eventApi = $this->getEventApi();
+        $event = $eventApi->getByFriendlyUrl($eventSlug);
+        $eventUri = $event->getUri();
+
+        $talkApi = $this->getTalkApi();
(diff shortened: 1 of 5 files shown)


app/src/Talk/TalkController.php
```diff
@@ -444,6 +446,34 @@ public function reportComment($eventSlug, $talkSlug, $commentHash)
         $this->application->redirect($url);
     }
 
+    public function unlinkSpeaker($eventSlug, $talkSlug, $username)
+    {
+        if (!isset($_SESSION['user'])) {
+            $thisUrl = $this->application->urlFor('talk', ['eventSlug' => $eventSlug, 'talkSlug' => $talkSlug]);
```
heiglandreas: I'd move this out of the if-clause to be able to use it as redirection URL after the action is finished.

app/src/Talk/TalkController.php
```diff
@@ -444,6 +446,34 @@ public function reportComment($eventSlug, $talkSlug, $commentHash)
...
+
+        $userApi = $this->getUserApi();
+        $user = $userApi->getUserByUsername($username);
+        $userId = $user->getId();
+
+        try {
+            // TODO: Make API call to unlink speaker from talk
+        } catch (Exception $e) {
+            // TODO: Handle exception
+        }
```
heiglandreas: What do you do after the action is finished?

I'd redirect to the talk-page (which should then have one speaker less)

railto: That's my plan, Just missed putting in a todo for it :)

app/src/Talk/TalkController.php
railto: Good point, updating now.

railto:
In commit/6196023 I mentioned that it's not showing flash message, just realised it is, but not at top of page.  My thought would be that this is outside the scope of this issue.  I can remove the flash message completely or leave it there until we decide if should be moved or not.

railto:
@liam-wiltshire @heiglandreas Could one of you take a final look at this when you get time please.

PR #387, opened 2016-11-12
JOINDIN-642

exussum12:
Added reject button
Added exception as a use statement

### DIFF
--- a/app/src/Talk/TalkController.php
+++ b/app/src/Talk/TalkController.php
@@ -16,7 +16,7 @@ use Symfony\Component\Form\FormError;
 
 class TalkController extends BaseController
 {
-    protected function defineRoutes(\Slim\Slim $app)
+    protected function defineRoutes(Slim $app)
     {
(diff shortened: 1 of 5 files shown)


app/src/Talk/TalkController.php
```diff
@@ -16,7 +16,7 @@
 
 class TalkController extends BaseController
 {
-    protected function defineRoutes(\Slim\Slim $app)
```
exussum12: imported on line 11.

akrabat:
okay to test

akrabat:
ok to test

exussum12:
build now passes

liam-wiltshire:
@exussum12 - Thanks for this, a feature we really needed. There is one minor UI change that'd be good if you get a chance - not a showstopper, hence I merged it anyway, but when you reject a talk claim, there is nothing in the UI that confirms this happens - if you look at approving a talk, the buttons get replaced with a 'Claim Approved' label - something similar for reject would be great if you get a minute :)

PR #390, opened 2016-12-04
Removing the setting of $result in places where it isn't used

skoop:
While researching another issue I noticed my PHPStorm complaining about unused variables. As it turns out, in quite a few places in the `UserController` the `$result` variable is being initialized with value `false`, but it is never used in the scope where it is defined. It is used later, but set to another value before usage. The initializing of `$result` to `false` therefore seems useless. In the spirit of leaving the code better than you found it, I decided to clean it up :)

### DIFF
--- a/app/src/User/UserController.php
+++ b/app/src/User/UserController.php
@@ -173,9 +173,8 @@ class UserController extends BaseController
         $token = $request->get('token');
         $userApi = $this->getUserApi();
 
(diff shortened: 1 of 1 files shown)


railto:
Ok to test

railto:
ok to test

liam-wiltshire:
ok to test

PR #391, opened 2016-12-08
Allows display of 3rd party apps

heiglandreas:
This PR adds an interface to list, show, edit and delete 3rd party apps.

As a user I can see my (and only my) apps that I registered for access to the joind.in API. This uses the API-changes from https://github.com/joindin/joindin-api/pull/383

This UI can be viewed via /client - There is currently no link or button available. Should this be merged we'd need to define where the interface should be implemented

### DIFF
--- a/app/src/Client/ClientApi.php
+++ b/app/src/Client/ClientApi.php
@@ -134,6 +134,4 @@ class ClientApi extends BaseApi
         return true;
 
     }
-
-
 }
--- a/app/src/Client/ClientApi.php
+++ b/app/src/Client/ClientApi.php
@@ -35,7 +35,7 @@ class ClientApi extends BaseApi
     /**
      * Get a specified client associated with the current user
      *
-     * @return array
+     * @return ClientEntity
      */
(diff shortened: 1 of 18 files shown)


app/src/Client/ClientApi.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Get all clients associated with the current user
+     *
+     * @return array
+     */
+    public function getCollection($queryParams)
+    {
+        $talks_uri = $this->baseApiUrl . '/v2.1/applications';
+
+        $talks = (array) json_decode(
```
topikito: What about passing the `json_decode` second argument `true` so it returns an array rather than an object?

heiglandreas: Doesn't work as the "internal" structure needs to be objects. Casting to an array only casts the outer object to an array while the internal objects still remain objects.

Would be a great idea but doesn't work in this special case without refactoring a large portion of the site. 😄

topikito: :+1: makes sense

PR #395, opened 2017-01-03
Moves Clients to users namespace

heiglandreas:
As we talked about the clients should also go into the users namespace.

So this PR handles that.


railto:
May need to rebase as getting conflicts on following files:

- app/src/Client/ClientController.php
- app/templates/Client/_common/client.html.twig
- app/templates/Client/details.html.twig

PR #396, opened 2017-01-05
Updates login so displays correct error message

railto:
Added in check to see if error message is Not verified and if so display the correct message
Part of #JOINDIN-718

### DIFF
--- a/app/src/User/AuthApi.php
+++ b/app/src/User/AuthApi.php
@@ -29,9 +29,9 @@ class AuthApi extends BaseApi
         if ($result) {
             $data = json_decode($result);
             if ($data) {
-                if (isset($data->access_token)) {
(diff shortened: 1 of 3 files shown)


app/src/User/AuthApi.php
```diff
@@ -29,9 +29,9 @@ public function login($username, $password, $clientId, $clientSecret)
         if ($result) {
             $data = json_decode($result);
             if ($data) {
-                if (isset($data->access_token)) {
+//                if (isset($data->access_token)) {
```
heiglandreas: I'd remove it completely. The history contains it anyway and when it's not needed you can just get rid of it.

railto: Crud, never realised I left that in, ty

PR #397, opened 2017-01-12
JOINDIN-642 Additional fix

exussum12:
@liam-wiltshire  This is the fix for rejecting a claim.

Also it looks like the looping was in the wrong place. Fixed that


liam-wiltshire:
Ok to test

liam-wiltshire:
ok to test

PR #399, opened 2017-01-21
Adds QR-Code links to event- and talk-details

heiglandreas:
This PR adds a link to an QR-Code to the event- and the talk-details. That way organizers and speakers can easily add a QR-Code to their slides for easier reference of the event or talk. That makes it
easier for attendees to find the talk or event and rate it.

### DIFF
--- a/app/src/View/Functions.php
+++ b/app/src/View/Functions.php
@@ -164,4 +164,16 @@ function initialize(Twig_Environment $env, Slim $app)
             ['is_safe' => ['html']]
         )
     );
+
+    /**
+     * Create a link to download a QR-Code for the given URL
+     */
+    $env->addFunction(
+        new Twig_SimpleFunction('qrcode', function ($url) use ($app){
+            return sprintf(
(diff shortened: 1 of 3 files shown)


railto:
Love the idea, however I'm not a massive fan of how the code is viewed, would it maybe be possible to have the QR code pop up in a modal instead of a full page reload that takes you off joind.in?  Would love others thoughts on this too.

heiglandreas:
The current implementation is "just" a copy of the existing implementation on the legacy site. 
bildschirmfoto 2017-01-31 um 09 28 21

That doesn't mean there's no room for improvement though 😄 

I also now see that the QR-code is only available on my own talks… Should that be that way? Or can anyone distribute a QR-code to your talk?

railto:
Ah, was not aware that was the case, if happy to go ahead then im happy to merge this in as long as we can get it updated at a later stage

railto:
My understanding is that anyone should be able to distribute a QR code for your talks.  In this PR it allows anyone to share the QR code including logged out which fits in line with how I'd personally like to see it, I don't see the point of having a link to generate a QR code but then forcing an event admin or speaker to get it themselves

PR #401, opened 2017-01-31
Offers to resend welcome email to un verified user upon login attempt

railto:
Offers to resend welcome email to un verified user upon login attempt

If a user that is not verified tries to login then this change will show a flash message to that effect and provide a link which the user can click in order to have the validation email resent, this does not resend the email automatically, rather the user needs to enter their email address.  I personally think this is a more secure way to go about it rather than having a single link click send the email.

Relates to #JOINDIN-718

### DIFF
--- a/app/src/User/AuthApi.php
+++ b/app/src/User/AuthApi.php
@@ -29,9 +29,7 @@ class AuthApi extends BaseApi
         if ($result) {
             $data = json_decode($result);
             if ($data) {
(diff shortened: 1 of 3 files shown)


railto:
Note that this is dependant on https://github.com/joindin/joindin-api/pull/389 to be merged first

railto:
Derp, I have already submitted this on https://github.com/joindin/joindin-web2/pull/396, closing this one as dupe

PR #406, opened 2017-03-14
Added the CTA to submit an event to the homepage

skoop:
I was talking with @yourmark of the WordPress Meetup Enschede. He is not a very regular user of joind.in and he mentioned that he missed a CTA to add an event on the homepage. I decided it was a good idea to get less experienced used on board and create a PR. I hope you like this ;)

### DIFF
--- a/app/templates/Application/index.html.twig
+++ b/app/templates/Application/index.html.twig
@@ -37,4 +37,13 @@
         <p>There are no events with an open call for papers.</p>
         {% endif %}
     </section>
+
+    <section>
+        <h3>Submit your event</h3>
+        <p>
+            Know of an event happening? Let us know! We love to get the word out about
+            events the community would be interested in and you can help us spread the word!
+        </p>
+        <a href="{{ urlFor('event-submit') }}" class="btn btn-primary col-xs-12">Submit</a>
+    </section>
 {% endblock %}


railto:
ok to test

heiglandreas:
Thanks @skoop for this great idea! 

Personally I'm pretty ambivalent on this feature. I love it to allow users to easily create an event that is not yet listed!

Currently though joind.in requires to have a person that is responsible for the event. So will that be the same person as the one that's reporting the event?  

And should it be possible for a user to create an event without being logged in? Or do we ask people to submit an event but - oh sorry, please first log in and - oh, you don't have an account yet, so please create one…? 

So on the one side, I like involvement of users, but on the other hand I see some impacts that need more thinking first. 

Perhaps it could be possible to show events every user can create without being logged in (still curated!) but then those events without a person in charge need to be claimed (and approved by an admin) before anything usefull can be done with it (including editing the event!)

That would require some changes to the API and Web2 though before we could merge this PR…

skoop:
@heiglandreas the only thing I've done is to take the block that's already on the event page, and I've copied that to the homepage. The simple reason for this is for people who use joind.in less than us to make it easier to find out how to add an event to the site. We should (imho) not change the fact that an event needs hosts, or that people should log in. That should all stay the same. The only thing I did was make it easier for people to find this feature.

heiglandreas:
Then I'm absolutely fine with that!

liam-wiltshire:
Totally agree with this - one of those things that's crossed my mind a number of times and I've never got round to sorting - thanks @skoop!

PR #408, opened 2017-03-22
Start to introduce SASS to the project

thatdamnqa:
Introducing SASS to the project. A couple of things to note here, that need to be done elsewhere in the project:
- The VM needs to have SASS installed
- The build process needs to run the new sass compile option in grunt
- We need to discuss/consider the affect this will have on contributors. Will this be pain free?

Right now, I've made no effort to split the SASS files into any meaningful files. If it gets merged into master, I'll start work on that.


composer.json
```diff
@@ -33,6 +33,7 @@
         "symfony/twig-bridge":   "~2.5",
         "org_heigl/daterange":   "~1.1",
         "guzzlehttp/guzzle":     "^6.2.1",
-        "monolog/monolog":       "^1.21"
+        "monolog/monolog":       "^1.21",
+        "twbs/bootstrap-sass": "^3.3"
```
heiglandreas: Can't we use bower or npm to include frontend-stuff instead of packagist? Feels somewhat strange…

web/css/main.css
```diff
@@ -0,0 +1,10 @@
+﻿/*!
```
heiglandreas: Should this file really be added to the repo? I could imagine that it will get really messy with different people committing this file…

And as it will be rendered newly during deployment the deployed file will *never* resemble the one checked in…

web/css/main.css.map
```diff
@@ -0,0 +1,7 @@
+{
```
heiglandreas: This should also not be checked in…

Perhaps add those two files to the .gitignore?

scss/README.md
```diff
@@ -0,0 +1,20 @@
...
+http://thesassway.com/beginner/how-to-structure-a-sass-project
+
+How to compile SASS
+==
+
+See https://www.jetbrains.com/help/phpstorm/2016.3/transpiling-sass-less-and-scss-to-css.html for instructions on how to compile SASS to CSS in phpstorm
+
+To compile the SASS into a CSS file using Grunt, run the following command:
+```
+grunt sass
```
heiglandreas: There is a phing-buildfile that should contain a task/target to do this compilation. At least as an add-on

composer.json
heiglandreas: as npm is already available and used for the jslinting I'd say we should definitely use npm to include frontend-stuff. So adding bootstrap via ```npm install bootstrap-sass -save``` should be ok…

app/templates/layout.html.twig
```diff
@@ -9,12 +9,11 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     {% block extra_meta %}{% endblock %}
-    <link href="/css/bootstrap.min.css" rel="stylesheet">
+    <link href="/css/main.css" rel="stylesheet">
     <link href="/css/bootstrap-theme.min.css" rel="stylesheet">
     <link href="/css/font-awesome-custom.css" rel="stylesheet">
```
heiglandreas: Can't we include this file via SCSS as well? The less files need to be loaded the better…

composer.json
Thatdamnqa: I'm fine with that. My concern was that it won't work with grunt that way: or am I wrong? (I'm no grunt expert)

web/css/main.css.map
Thatdamnqa: The reason I checked it in is so that it works out of the box on checkout/install. But I'm not adverse to leaving it out (for the reasons you gave)

scss/README.md
Thatdamnqa: Any chance you can work with me to get that working?

app/templates/layout.html.twig
Thatdamnqa: Yes. I was planning on doing that around the same time as splitting the css file. I wanted to just get a proof of concept out there :)

Gruntfile.js
```diff
@@ -15,8 +16,21 @@ module.exports = function(grunt) {
...
+            'loadPath': 'sass',
+            'style': 'compressed'
+          },
+          dist: {
+              files: [{
+                expand: true,
+                cwd: 'scss',
+                src: ['main.scss'],
+                dest: 'web/css/',
+                ext: '.css'
```
exussum12: can this generated file be added to the git ignore please

exussum12:
This adds a load of CSS but doesnt remove any ?

heiglandreas:
@exussum12 The (now created) CSS was left in - see https://github.com/joindin/joindin-web2/pull/408#discussion_r107558491

exussum12:
@heiglandreas Sorry was refering to  "* This file is the original joindin.css file."  and no removals meaning there are now 2 versions of that code. one in the scss file and one in its original place

heiglandreas:
Which makes no sense at all IMHO. Both should be removed in my eyes… Thanks for spotting!

heiglandreas:
@markcotton Would you mind working with @Thatdamnqa together to bring #367 into this PR? That'd be great as both try to bring SASS to joind.in.

thatdamnqa:
I've split the file into modules as suggested by @marktcotton. 

Further refinements to come as future PRs.

railto:
@thatdamnqa can you have a look at the merge conflict in app/templates/layout.html.twig please

svpernova09:
Closing as abandoned by the author. Feel free to update and reopen.

PR #412, opened 2017-05-26
Refactored controllers and services to use built in Slim DI container

aenglander:
Resolves JOINDIN-711


liam-wiltshire:
ok to test

liam-wiltshire:
It looks good to me, however with it being a fairly big change, @railto have you got a few minutes to give it a second set of eyes please?

PR #416, opened 2017-06-29
Title should be escaped, not raw

giorgiosironi:
We don't escape the title elsewhere, only do some filtering that strips away tags in the API writes. Therefore, escape it here so that values like `The wonder of <blink>` are allowed


liam-wiltshire:
ok to test

PR #417, opened 2017-07-02
Speakers can not change speaker info after start of talk (JOINDIN-749)

wimg:
Not allowing speakers to change speaker information after the talk has been started. See JOINDIN-749


app/src/Talk/TalkController.php
```diff
@@ -475,15 +480,21 @@ public function unlinkSpeaker($eventSlug, $talkSlug, $username)
 
         $unlinkSpeakerUri = $talkUri . "/speakers/" . $userId;
 
-        try {
-            $talkApi->unlinkVerifiedSpeakerFromTalk($unlinkSpeakerUri);
-        } catch (Exception $e) {
-            $this->application->flash('error', $e->getMessage());
+        $isAdmin = $event->getCanEdit();
+        if (!$isAdmin && $talk->getStartDateTime()->format('U') > time()) {
```
heiglandreas: Please compare ```$talk->GetStartDateTime()``` to ```new DateTimeImmutable()```

app/src/Talk/TalkController.php
```diff
@@ -170,6 +170,11 @@ public function editTalk($eventSlug, $talkSlug)
             if ($form->isValid()) {
                 $values = $form->getdata();
 
+                // not allowing speakers to remove themselves after a talk has started (JOINDIN-749)
+                if (!$isAdmin && $data['start_date']->format('U') < time()) {
```
heiglandreas: Please compare ``` $data['start_date']``` to ```new DateTimeImmutable()```

app/src/Talk/TalkController.php
```diff
@@ -475,15 +480,21 @@ public function unlinkSpeaker($eventSlug, $talkSlug, $username)
...
+        if (!$isAdmin && $talk->getStartDateTime()->format('U') > time()) {
+            try {
+                $talkApi->unlinkVerifiedSpeakerFromTalk($unlinkSpeakerUri);
+            } catch (Exception $e) {
+                $this->application->flash('error', $e->getMessage());
+                $this->application->redirect($url);
+            }
+
+            $this->application->flash('message', 'Speaker has been removed from this talk.');
+            $this->application->redirect($url);
```
heiglandreas: Can you add a ```return``` here to stop execution of the method in any case?

app/src/Talk/TalkController.php
```diff
@@ -475,15 +480,21 @@ public function unlinkSpeaker($eventSlug, $talkSlug, $username)
...
-            $talkApi->unlinkVerifiedSpeakerFromTalk($unlinkSpeakerUri);
-        } catch (Exception $e) {
-            $this->application->flash('error', $e->getMessage());
+        $isAdmin = $event->getCanEdit();
+        if (!$isAdmin && $talk->getStartDateTime()->format('U') > time()) {
+            try {
+                $talkApi->unlinkVerifiedSpeakerFromTalk($unlinkSpeakerUri);
+            } catch (Exception $e) {
+                $this->application->flash('error', $e->getMessage());
+                $this->application->redirect($url);
```
heiglandreas: Can you add a ```return``` here to stop execution of the method in any case?

app/src/Talk/TalkController.php
```diff
@@ -475,15 +480,21 @@ public function unlinkSpeaker($eventSlug, $talkSlug, $username)
...
+            }
+
+            $this->application->flash('message', 'Speaker has been removed from this talk.');
+            $this->application->redirect($url);
+        } else {
```
heiglandreas: When you added a return in the line above this ```else``` is not needed anymore and the content of the ```else```-block can move up one indentation level.

app/src/Talk/TalkController.php
exussum12: The redirect throws a `Stop` Exception so may not be needed, though returning the value makes it more clear

app/src/Talk/TalkController.php
exussum12: The check is made twice also (differnt sign), can a variable be used instead ? `$eventHasStarted = $talk->getStartDateTime()->format('U') > time()` then check is `!$isAdmin && $eventHasStarted` and the other check is `!isAdmin && !eventHasStarted`

app/src/Talk/TalkController.php
heiglandreas: For me it's a matter of clearness :wink:

app/src/Talk/TalkController.php
heiglandreas: Those are two separate methods, so IMHO that's OK here…

app/src/Talk/TalkController.php
wimg: Since this is a one-or-the-other (true boolean) situation, I believe an if-else is the correct usage here. If the _if_ was unrelated to the outcome of the contents of the else, things would be different. In this case the _if_ is exactly about the message in the else.
I could reverse the logic if you feel that's better ?

wimg:
Made the requested changes (except for the last return - see comment above)

app/src/Talk/TalkController.php
heiglandreas: Somehow I actually do :smile:

Personally I feel it'd be "cleaner" to have the "intended" flow of the function in the first indentation-level. For this function that would mean actually unlinking the user. So when the user is not able to unlink that's not the intended flow and would therefore happen within an ```if```-block. But that's my personal thing and I'm absolutely happy with the functionality being available at all ;)

liam-wiltshire:
ok to test

PR #418, opened 2017-10-03
Show only starred talks on event schedule page

mheap:
This PR adds the ability to show only starred talks. If there are no starred talks in a certain time slot, that time is no longer shown.

You can switch between list/grid and starred/unstarred without losing your selection. We don't currently set a cookie for starred/unstarred like we do for layout, but could do if required. Personally, I think seeing all talks at once is more useful.

I've attached screenshots of how this looks. This is my first contribution in a long time so please do review closely and let me know if there are any changes to be made.

### List view
screen shot 2017-10-03 at 21 02 39

### Grid view
screen shot 2017-10-03 at 21 07 15


railto:
Thanks for the PR Michael.

I haven’t fully reviewed the code yet, but am unsure of the functionality.  If it does not display any talks for a time slot that does not have a starred talk, then how are you to see the talks in that slot to be able to star them?

Maybe I’m thinking about this wrong but that seems like a pretty big issue to me.

Thanks again,

Mark

mheap:
Sorry, I wasn’t clear with my explanation. By default all sessions are shown. The behaviour
I described occurs when you select “Starred only” (see links on screenshot)

liam-wiltshire:
Thanks for the PR @mheap  - this is a feature I've wanted for ages :D 

One thing I have noticed - from the 'base' event URL - e.g. http://dev.joind.in/event/php-meetup you see the schedule tab by default, which means you can still click on the 'Starred Only' button, however that then produces a URL such as http://dev.joind.in/event/php-meetup/starred which results in a 404.

Suggest that you check the currently URL, and if it's the base URL for the event you force the 'Starred Only' button to go to /schedule/list/starred ?

Thanks!

mheap:
@liam-wiltshire All done! I've checked if the URL ends in `/schedule/list` or `/schedule/grid` and behave accordingly.

Though this fixes the issue, I'm not sure it's the correct fix. The event default page calls the correct action (https://github.com/joindin/joindin-web2/blob/master/app/src/Event/EventController.php#L178-L198), where as the schedule default page redirects (https://github.com/joindin/joindin-web2/blob/master/app/src/Event/EventController.php#L308-L319).

Whilst I'm happy for this to be merged as-is, it may be worth revisiting the delegation/redirection inconsistency and deciding which behaviour is preferable.

PR #419, opened 2017-10-15
Enable github login

exussum12:
Adds a button and the call to the API to enable github users to register
or login without an account here

### DIFF
--- a/app/src/User/AuthApi.php
+++ b/app/src/User/AuthApi.php
@@ -115,6 +115,36 @@ class AuthApi extends BaseApi
                 }
             }
         }
+
+        return false;
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


liam-wiltshire:
ok to test

liam-wiltshire:
Thanks for this @exussum12 - I'm waiting for https://github.com/joindin/joindin-api/pull/424 to be sorted before I merge anything else that relies on the API as that PR touches lots of files, but as soon as that's through I'll look at this

exussum12:
I am going to update this to be id based instead of email so will need a little extra time either way

svpernova09:
Closing as we've also put on hold adding providers until we clean up the oath code we're using

PR #422, opened 2017-10-22
JOINDIN-750 #close

cal-tec:
Fix issue with disappearing talks in schedule view that share the same name. Reported on Ji raas JOINDIN-750


railto:
thanks @cal-tec

PR #423, opened 2017-10-22
Add target _blank to href tags for QR code links

cal-tec:
I notice whist browsing the site that the QR-Code pages provide no way to intuitively go back to the site. For this reason I inadvertently closed the tab that I was using to browse the site instead of pressing back.

I feel this PR is an improvement to the current system as it adds the functionality to open the QR-Code's in new tabs.


railto:
I would tend to agree that having a notice that this would open in a new tab/window would be preferable for this change

railto:
@cal-tec, I'd love to get this merged in, would you mind adding the notice? Once that's added I'll look to getting it merged straight away.

cal-tec:
Added (opens in new window) warning after QR-Code Links as suggested 😃

railto:
Thanks @cal-tec

PR #424, opened 2017-10-22
Alter  TalkMediaFormType to enable url validation

cal-tec:
Media links are not validated. This results in relative links if a user does not bother to enter a protocol.

I understand this may be restrictive in someways as users can enter related talk links but I believe that for these to work effectively throughout the site they must also be fully qualified URLS.

This PR adds validation to the form via the  TalkMediaFormType Class.


heiglandreas:
Thanks for this PR. Did you think about adding that validation to the API as well? Otherwise people might be able to still add relative links directly via the API…

cal-tec:
I will have a look and see what I can do.

svpernova09:
Thanks!

PR #426, opened 2017-11-03
Fix Talk Links

### LINKED ISSUE #420: Add a 'delete' button when adding media to a talk
Currently, if you click the Add Media button too many times, there is no way to remove the additional fields that are created - if tyou don't provide a value for these fields, an error then results.

Have raised an issue on the API side https://github.com/joindin/joindin-api/issues/436 , but would also be good to fix on the web side as well.

exussum12:
Fixes #420 

Doesnt add delete button, Dont have time to add that currently but will try and update that!


railto:
ok to test

railto:
Thanks @exussum12

PR #427, opened 2017-11-05
Adds new rating-icon

heiglandreas:
This commit adds a new rating icon that allows to also show the rating of the currently logged in user. So users can see immediately whether they did rate a talk or not and how their rating was compared to the current average one.

The new rating Icons would look something like this:

bildschirmfoto 2017-11-05 um 19 35 44

This addresses https://joindin.jira.com/browse/JOINDIN-759


railto:
Sorry for delay in looking at this, looks fantastic and love the idea, only thing I have, is it possible to vertically center align the thumb icon inside the bubble? (For perspective, I've tested on firefox and chrome on ubuntu and firefox on macos)

Otherwise it's a great change and makes a lot of sense

michaelcullum:
@heiglandreas Want to resolve the conflict here so we can get this in?

heiglandreas:
No iOS, I did *not* want to close this issue!

On the contrary! I *do* want to resolve the conflicts and get on with it!!

heiglandreas:
Rebased and resolved the conflicts.

app/templates/Event/schedule-list.html.twig
```diff
@@ -50,7 +50,7 @@
                         </td>
                         <td class="agenda-cell">
                             <div class="talk-info">
-                                {% include '_common/rating.html.twig' with {'rating': talk.getAverageRating, 'style': 'vertical', 'size': 'small', 'include_rating_placeholder': true} %}
+                                {% include '_common/rating.html.twig' with {'rating': talk.getAverageRating, 'userrating' : talk.getUserRating, 'style': 'vertical', 'size': 'small', 'include_rating_placeholder': true} %}
```
mcneely: I realize this is already established and I'm not going to request you to make changes for this PR but it would make sense for us to start implementing Twig Macros for this sort of thing.

PR #435, opened 2018-03-24
Json Decode with Big Ints

### LINKED ISSUE #433: "Page not found" when clicking on my name
I apologize in advance if this isn't the correct venue for this but when I click on my name in the menu I receive an error that the page isn't found.  The URL shows up as https://joind.in/user/1.0155671600264E+16

exussum12:
Fixes #433


exussum12:
There is a api version of this too. I think it fixes all of the instances where this happens

railto:
Thanks @exussum12

PR #439, opened 2018-04-16
Add a privacy policy to the joind.in-Website

heiglandreas:
This is a first draft and thought as base for a discussion.

I generated this privacy policy using https://dsgvo-muster-datenschutzerklaerung.dg-datenschutz.de/

I'm sure there are some alterations necessary especially as joind.in is a community-project and does not have a CEO or similar. 

I used my personal address for contacts as there is a postal address necessary but I'd rather say we use the address of @liam-wiltshire as he's the one _in charge_ ;-)

And I'm not sure whether we should add this as separate page or use a controller to handle static pages like that one.


liam-wiltshire:
Thanks Andreas ;). I was actually talking to our work legal about this a
couple of weeks ago - not that I'd done much about it since. There are a
couple of things we need to do, although our limited processing of PII  (I.
E. We don't do anything with it!) means it's much more straightforward than
for most!

heiglandreas:
What concerns me more is that that little data we have can be retrieved by everyone via the API. Therefore we can't tell users where their data went to. 

I think we need to make it more explicit that their data can and will be retrieved by third parties outside our reach.

heiglandreas:
It is "only" the username, the real name and - if available - twitterhandle. but these are PIIs and therefore fall under the GDPR...

web/privacy.html
```diff
@@ -0,0 +1,521 @@
...
+
+<h4>2. Name and Address of the controller</h4>
+<p>Controller for the purposes of the General Data Protection Regulation (GDPR),
+    other data protection laws applicable in Member states of the European
+    Union and other provisions related to data protection is:
+
+</p>
+
+<address>joind.in
+<br/>Forsthausstraße 7
```
heiglandreas: This address should be replaced with someone elses ;)

web/privacy.html
```diff
@@ -0,0 +1,521 @@
...
+    </li>
+    <li><h4>i) Right to withdraw data protection consent </h4>
+        <p>Each data subject shall have the right granted by the European legislator to withdraw his or her consent to processing of his or her personal data at any time. </p>
+
+        <p>f the data subject wishes to exercise the right to withdraw the consent, he or she may, at any time, contact any employee of the joind.in.</p>
+
+    </li>
+</ul>
+<h4>9. Data protection provisions about the application and use of Facebook</h4>
+<p>On this website, the controller has integrated components of the enterprise
```
heiglandreas: We don't have a "share on facebook" button only the possibility to log in via facebook. So this will need to be rewritten…

web/privacy.html
```diff
@@ -0,0 +1,521 @@
...
+    https://facebook.com/about/privacy/, provides information about the
+    collection, processing and use of personal data by Facebook. In addition,
+    it is explained there what setting options Facebook offers to protect the
+    privacy of the data subject. In addition, different configuration options
+    are made available to allow the elimination of data transmission to
+    Facebook. These applications may be used by the data subject to eliminate
+    a data transmission to Facebook.</p>
+
+<h4>10. Data protection provisions about the application and use of Twitter</h4>
+<p>On this website, the controller has integrated components of Twitter.
```
heiglandreas: We don'T have a "share on twitter" button, only the possiblity to log in using twitter.

mcneely:
@heiglandreas Any chance you can reformat this to MD file and just push it up to the docs repo?

iansltx:
Dunno about "just". It should be on the website, and is specific to the website rather than the project as a whole.

iansltx:
Thinking about this a bit more, I'm a bit further convinced that this should be part of the web2 repo, mainly because the alternative would be linking to docs.joind.in from an install that technically could be not the centralized one. Including the privacy policy in the web2 repo means that such an install could fork the project and change the privacy policy to one they're willing to be bound to, which is what you want when the repo could theoretically be forked by evil-talk-ratings.rocks or something :)

Open to counterpoints here since, yes, evil-talk-ratings.rocks could swap the link for their own privacy policy link. Or we assume that nobody installs this other than the canonical site.

mcneely:
Yeah, let's go ahead and get this merged. Moving this to docs would make it easier to edit but at this juncture, we need a privacy policy in place due to issues with FB login and it's not worth a debate.

heiglandreas:
Then I'd say let's get this merged to have it for FaceBooks sake and then adapt it afterwards as appropriate.

F.E. the address is currently my one (which I'd not want to have in there for too long) but as Joe gets the NonProfit up and running it should be replaced with that address....

PR #443, opened 2018-07-03
Allow users to add a small biography to their profile

Pton:
This will allow a user to add some more information regarding their occupation, specialization, interests and experience. That additional information can be useful for users interested in a speaker, for event organizers to get a feel for a speaker or for getting some context behind a comment.

This PR depends on changes in the API made in joindin/joindin-api#449


railto:
Thanks for adding this in

Issue #444, opened 2018-07-05
Feature to search for users


Pton:
Currently users can search for events and talks. I would like to be able to add the possibility to search for users to that. This will allow users to be able to look up talks for a speaker they notice on twitter and it will allow event organisers to be able to quickly navigate to the user profile of a speaker in order to see previous talks or events. My employer (@IbuildingsNL) has given me some time to implement this feature.

I'm happy to write the implementation of this feature, but I'm unsure about the design of the search page.  Adding a third column to the search page reduces the space each column gets on desktop and tablet views, making the summaries quite small. So far I've come up with two solutions:
1. Adding a third column for users without adjusting any other styles
![design1](https://user-images.githubusercontent.com/10808712/42332353-0b3d0c78-8078-11e8-8bff-2d8a0539ff25.png)
2. Adding a third column for users and adjusting the event summary style slightly
![design2](https://user-images.githubusercontent.com/10808712/42332354-0b5c8512-8078-11e8-939b-583e968d71bd.png)

I would like to know what you think of this feature in general and what design solution you think is best. Any other ideas or suggestions are appreciated as well 😄

railto:
Closing as I have now merged the PR's that you opened for this.  Thank you for your contribution

PR #445, opened 2018-07-10
Expand website search with ability to search for users

Pton:
As described in #444 I'd like to add the ability to search for users. This will allow for improved search on the website. For instance a user will be able to look up talks for a speaker they notice on twitter. Additionally event organizers will be able to quickly navigate to the user profile of a speaker in order to see previous their talks or attended events.

With this PR also affect the search page design for the event summary slightly (see option 2 in #444).

This PR depends on changes in the API made in joindin/joindin-api#451


railto:
Thank you for the path, I really appreciate it

PR #446, opened 2018-07-24
Add Docker Support

exussum12:
Need to over ride redis and set default timezone

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -6,6 +6,9 @@
 if (in_array(substr($_SERVER['REQUEST_URI'], -4), ['.css', '.jpg', '.png'])) {
 	return false;
 }
+if(!ini_get('date.timezone')) {
+    date_default_timezone_set('UTC');
+}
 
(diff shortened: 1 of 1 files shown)


railto:
Thank's for adding this @exussum12, quick question tho, considering we have the VM project for providing a production'ish environment, what benefit will this bring? My main concern here being that we'll be providing 2 different development environments that we'd then need to support.

exussum12:
Docker is much quicker for linux based envs, just allowing extra choice

EricHogue:
This PR looks good. But as for adding Docker, I think we should either migrate to Docker or stay with Vagrant. 

If we try to keep both, it means more maintenance work. It might also be overwhelming for new contributors who won't know which solution to pick.

mcneely:
As this has no conflicts and doesn't break anything just makes the code usable for those wishing to run and test under docker, there's no reason to not merge. Doing so now.

PR #448, opened 2018-10-04
Removes the Apps page

railto:
Mobile apps are now deprecated, this removes the page
that links to the iOS and Android apps.


liam-wiltshire:
@railto, can we change this to have a page stating the apps have been discontinued please? Just in case there are rogue links on the web we don't know about. :)

PR #449, opened 2018-10-07
Update the default session time

exussum12:
Its currently a session cookie, this will make it a week long. This will stop logging people out


railto:
Thanks @exussum12

Issue #440, opened 2018-05-30
Feature request: Include talk descriptions on talk list page


mkopinsky:
It would be really cool if there were a way to include talk descriptions as part of the page, so I could print a list of talks with the descriptions. Sometimes at a conference I really like to be tech-free so I can pay attention to the talks, and that printable view would help me choose talks from just a piece of paper.

Thanks for the great site!

czmole:
Hi, 

You mean you're printing the pages like this https://joind.in/event/dutch-php-conference-2018 ?
Or there is some other page you have in mind?

Regards,
Cata

mkopinsky:
I mean that on that page, I should be able to toggle to see the talk descriptions e.g. from https://joind.in/event/dutch-php-conference-2018/making-games-with-reactjs

e.g. something like:
![image](https://user-images.githubusercontent.com/591435/40883441-f70ec5fa-66ca-11e8-88b1-b98fd32420e8.png)
(I think toggling via a checkbox would probably be toggling with another button group; I just added another button group because it was easier to mock up.)

czmole:
Thanks, it's just what I had in mind also - but wanted to be sure I get it right

czmole:
I am picking this.

railto:
This has now been merged, thank you both

PR #452, opened 2018-10-16
Feature request: Include talk descriptions on talk list page #440

czmole:
Added a new button group for all lists (it does not apply to the grid as I thought it feels too crowdy). I choose the button group because I think is more expressive and integrates better with that "toolbar" of button groups. 

An alternative would be I think just a checkbox that would word something like "Include talks descriptions". 

What do you think?


railto:
Hi @czmole,

Thanks for taking the time to create this.

close #440

mkopinsky:
Cool! Thanks!

czmole:
Thank you also!

PR #730, opened 2018-10-28
TIDY CODE: Remove duplicate code and methods and consolidate method calls

cal-tec:
Tidy code remove pointless TalkEntity->getCommentUri and update method calls (well just 1) to use TalkEntity->getCommentsUri Method


svpernova09:
Thanks!

PR #732, opened 2018-10-28
Remove pointless (legacy?) code to separately lookup username on TalkComments

cal-tec:
Remove code to separately lookup username on TalkComments as the users username is provided by the API on a TalkComment. I presume this is legacy code from a point in time where the API did not return usernames on Talk Comments.


svpernova09:
Thanks!

PR #733, opened 2018-10-29
Enable Event Comment Links to users

cal-tec:
Currently, the Twig template for comments (used by both Talk and Event Comments) provides for linking to user profiles from comments if the relevant data is provided to the template. This feature is implemented for TalkComments but not for EventComments. This PR enables the EventCommentEntity to access the getUsername method. The reason for this is to future proof web2 to allow for when the API provides a users username for EventComments as it currently does for TalkComments. This PR is in preparation for [my other PR](https://github.com/joindin/joindin-api/pull/611) on the API but does not break web2 so there should be no reason not to merge it prior to any merge of the API PR


svpernova09:
Thanks!

PR #736, opened 2019-01-14
Update CONTRIBUTING.md

exussum12:


CONTRIBUTING.md
```diff
@@ -29,31 +29,10 @@ the mailing list.
...
-```
-
 ## PHP Version
 
 The current PHP version for joind.in is PHP 5.6.
 
 ## Code Style
 
-All PHP code follows [psr-2](http://www.php-fig.org/psr/psr-2/). One way to help ensure that your submitted code is PSR-2 is to run [PHP Code Sniffer]() against PSR-2 and the [.editorconfig](http://editorconfig.org/) plugin for your editor before submitting.
+All PHP code follows [psr-2](http://www.php-fig.org/psr/psr-2/). One way to help ensure that your submitted code is PSR-2 is to run [PHP Code Sniffer](https://github.com/squizlabs/PHP_CodeSniffer) against PSR-2 and the [.editorconfig](http://editorconfig.org/) plugin for your editor before submitting.
```
heiglandreas: Shouldn't we give the exact commandline for easier reference here? It should be something like `phpcs --standard psr2 .` so people know exactly what to do? Alternatively add a composer-script that does exactly that so people can "simply" run `composer checkstyle` and everything else is taken care of. Perhaps including downloading the phpcs-phar-file...

CONTRIBUTING.md
```diff
@@ -16,11 +16,11 @@ read/subscribe to the following resources:
 ## What to work on
 
 You should work on what you want to. Our bug tracker is
-here: https://joindin.jira.com
+here: https://github.com/joindin/joindin-web2/issues
 
 Any issues that have the "hackathon" or "easypick" label are ones that we think
-are a good starting point. This [JIRA filter](https://joindin.jira.com/issues/?jql=project%20%3D%20JOINDIN%20AND%20labels%20in%20(hackathon%2C%20%22OR%22%2C%20easypick)) will give you the list of all
-current issues with the "hackathon" or "easypick" issues.
+are a good starting point. This [filter](https://github.com/joindin/joindin-web2/issues?q=is%3Aopen+is%3Aissue+label%3Aeasypick) will give you the list of all
```
heiglandreas: While the jira-link referenced *all* the easypicks for the joind.in-project this now references only the easy-picks for web2. Do we want to have easypicks for the whole project?

heiglandreas: https://github.com/orgs/joindin/projects/1

exussum12: Says didn't find anything here for me ?

PR #737, opened 2019-01-19
Stop the gc logging out

exussum12:
Note, This will massively increase the amount of disk space sessions take up. Though I still can't imagine that being > 2 GB or so


exussum12:
This would be better as a ini change, depending on how php is set up on the server

exussum12:
Ignore this for now, a better fix is incoming. We shouldnt need to relay on the session

exussum12:
This pr will fix the issue. 

The better fix is to remove sessions as the storage method. So working on that but it's a bigger job than I thought.

exussum12:
Ok speaking to Ian about this. There is not a large amount of disk space left on the server. This fix will likely fill that as the session contains the entire user object. 

The better fix here is to store the token in a cookie and only cache the user object in the session. This can be cleared with the current GC after 25 mins. 

If the session is empty we will need to re populate the session with the user object. Currently there is not a way to do this. So we need to add either and endpoint to get user from token. Or a way to store and authenticate the user from web 2 ( would mean adding something like libsodium to be completely safe. Slims encrypted cookies may be good enough though here ) in cookies. 

Both ways are not a quick fix 

If we can easily get more disk space easily  the quick and easy way is to change this ini setting

mcneely:
Just curious but why do we have week-long user sessions?

exussum12:
We don't. This suggests that we should. It logs out every 24 mins which is annoying. 

Ideally there shouldn't be a session at all but that's a bigger fix

svpernova09:
Closing this as I believe based on the discussion in https://github.com/joindin/joindin-web2/issues/834 we're going to change this at the host level via Platform.sh configuration.

PR #746, opened 2019-02-02
add Travis

exussum12:


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: php
+dist: trusty
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+    - $HOME/.cache/composer/files
+matrix:
+  include:
+    - php: 5.6
```
michaelcullum:
```suggestion
    - php: 5.6
    - php: 7.3
  allow_failures:
    - php: 7.3
```

svpernova09: IMHO we should not let 7.3 fail currently as we need to know it's status since we should be targeting 7.3 with the new server/setup. We should add & allow `nightly` to fail.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  fast_finish: true
+before_script:
+  - composer global require squizlabs/php_codesniffer 
+  - composer global require phpunit/phpunit
```
michaelcullum: These should really be composer dev dependencies (at least phpunit) and we should adjust build.sh to use `./vendor/bin/phpunit` but happy to leave that for a separate PR

.travis.yml
michaelcullum: 7.3 will fail until compatibility is finished though and we shouldn't block merging all PRs until then. Setting off on a note of it being okay to merge failing PRs is non-ideal I think.

We would then remove `allow_failures: 7.3` in the compatibility PR (and remove 5.6).

svpernova09: Sounds good!

.travis.yml
iansltx: +1 for separate PR, after #750 lands

PR #750, opened 2019-02-04
Composer updates & PHP 7 Compat

michaelcullum:
- [x] Remove vendor and add to gitignore
- [x] Update symfony deps to a supported LTS (2.8 for now)
- [x] Travis PHP 7 testing
- [x] Add composer step to jenkins
- [x] Add phpunit and other development deps
- [x] Adjust build.sh to use require-dev deps

For reviewing this PR, look at the individual commits as the diff is too big with removing `vendor/*`


michaelcullum:
Requires #746

iansltx:
@dstockto we're going to want to run a release when this gets merged, so that if this causes failures/explosions we catch them now. Let's figure out a time in chat.

michaelcullum:
> For example, I've assumed composer and php are available and in the path, that's probably not true. If it's not true we'll add a phar to the repository or fetch it with cURL but I thought I'd try doing it the nicer way first.

I've just checked this and both php and composer are available in the PATH and I've just updated composer to the latest version.

ericpoe:
Looks great!

PR #756, opened 2019-02-06
Allow users to delete their own accounts

iansltx:
We have the API endpoint for this. Includes revised wording to indicate when you're deleting your own account.

**NOTE:** Untested! If someone can pull this down and confirm functionality, that'd be amazing. Otherwise I'll do so on Saturday or so.


iansltx:
Per https://github.com/joindin/joindin-api/issues/621, closing this one.

PR #764, opened 2019-02-28
Fix for platform.sh redis.

mcneely:
Moves default redis server to the config so it can be configured in platform.sh

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -104,7 +104,6 @@ $app->container->singleton(\Application\CacheService::class, function ($containe
 
     $redis = $container->settings['custom']['redis'];
     $keyPrefix = $redis['keyPrefix'];
-    unset($redis['keyPrefix']);
 
     if ($host = getenv('REDIS_HOST')) {
         $redis = "tcp://$host:6379";
--- a/web/index.php
+++ b/web/index.php
@@ -103,17 +103,14 @@ $app->container->set('access_token', isset($_SESSION['access_token']) ? $_SESSIO
 $app->container->singleton(\Application\CacheService::class, function ($container) {
 
     $redis = $container->settings['custom']['redis'];
-
-    if ($host = getenv('REDIS_HOST')) {
-        $redis['scheme'] = 'tcp';
-        $redis['host'] = $host;
-        $redis['port'] = 6379;
-    }
-
     $keyPrefix = $redis['keyPrefix'];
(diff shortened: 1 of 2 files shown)


web/index.php
```diff
@@ -101,13 +101,19 @@
...
 
 $app->container->singleton(\Application\CacheService::class, function ($container) {
-    $redisServer = null;
+
+    $redis = $container->settings['custom']['redis'];
+
     if ($host = getenv('REDIS_HOST')) {
-        $redisServer = "tcp://$host:6379";
+        $redis['scheme'] = 'tcp';
+        $redis['host'] = $host;
```
heiglandreas: Why not set the host only here? After all the others are already preset, aren't they?

mcneely: On platform we're just copying over the config file and it pulls in special variables.

heiglandreas: Find. But if we only set the host here there should only be the host variable set in this if statement. The port and the scheme should not be set here.

zghosts: I agree, setting scheme and port here is redunant

web/index.php
```diff
@@ -101,13 +101,19 @@
...
+        $redis['port'] = 6379;
     }
 
-    $client = new Predis\Client($redisServer);
-    $keyPrefix = $container->settings['custom']['redisKeyPrefix'];
+    $keyPrefix = $redis['keyPrefix'];
+    unset($redis['keyPrefix']);
```
zghosts: Unsetting $redis['keyPrefix'] here feels wrong is we have to unset it here, we shouldn't set it in this array in the first place

web/index.php
zghosts: Never mind my remark, for breaking changes sake setting scheme and port should be done.

web/index.php
mcneely: I went ahead and removed the unset. I don't think it's going to hurt anything anyway.

web/index.php
mcneely: redis config can be set several ways from the config either string or array and can be TCP or Unix sockets.  If it's changed in the dist I didn't want that to effect REDIS_HOST if that is the override. I put the host back to being a string. For example if we set in the dist to use Unix sockets instead of TCP,  scheme, and port wouldn't be available properly if someone was using docker and overrides it.

PR #776, opened 2019-03-06
makes namespaces psr4 compliant

zghosts:


app/src/Event/EventFormType.php
```diff
@@ -1,12 +1,14 @@
 <?php
 
-namespace Event;
+namespace JoindIn\Web\Event;
 
+use Event\Constraint;
```
ericpoe: Should this be `use JoindIn\Web\Event\Constraint;`?

zghosts: actually it's redundant

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
ericpoe: Do we want to have IDE files/dirs in the local `.gitignore` instead of leaving it up to the user to add them to their global gitignore?

zghosts: If the local user hasn't added it to their global gitignore then we risk having it end up in the repo, better safe than sorry as it doesn't hurt

composer.json
```diff
@@ -36,11 +29,12 @@
...
         "symfony/security-csrf": "~2.8",
         "symfony/twig-bridge":   "~2.8",
-        "twig/twig":             "~1.15"
+        "twig/twig":             "~1.15",
+        "ext-json":              "*"
     },
     "require-dev": {
         "jakub-onderka/php-parallel-lint": "^1.0",
-        "phpunit/phpunit": "^8.0",
+        "phpunit/phpunit": "^7.0",
```
iansltx: Why did we downgrade phpunit here?

iansltx:
@zghosts Cleaned up a couple minor things that Storm flagged as I was cross-checking changes there. If you'd rather back those out, feel free to rebase 'em back out and I'll submit a PR for them separately.

Beyond that, my only question is why we're bumping PHPUnit versions down. Otherwise, lgtm, though I'd like to be clear of major confs before we deploy, just in case.

composer.json
zghosts: symfony formtype tests are failing because of strict types in phpunit 8.0, I can upgrade the components we use to 3.4, which would be out of scope for this pr, but it would solve the phpunit problems.

zghosts:
Adding the changes to this pr is fine by me, we should deploy to staging first and test the crap out of it before we deploy to production.

afilina:
This looks like a stale PR. I would recommend closing it and revisit the topic at a later time.

iansltx:
This looks like it incorporates the Symfony 3.4 change, and maybe dep updates, so we'll want to merge this after those two land. Maybe that could happen next week, but that depends on re-rollls of the other two. My guess is it wouldn't take a *ton* of work since the codebase basically hasn't changed in quite awhile.

sandacode:
I was in contact with Andreas about doing some upgrading work on the codebase. I'll be covering some of the same stuff you two did, and will probably borrow a few of your commits in the course of that work.

PR #786, opened 2019-03-25
Add honeypot for contact form

heiglandreas:
This adds a Form-Field that has to be empty. A so called HoneyPot. A
trap for bots that usually fill out every form field that seems
relevant. As the introduced field needs to be empty every
form-submission that contains that field will automatically be filtered
out as the form wasn't valid.

### DIFF
--- a/app/templates/Application/contact.html.twig
+++ b/app/templates/Application/contact.html.twig
@@ -36,6 +36,13 @@
             {{ form_row(form.comment) }}
         </fieldset>
     </div>
+    <div class="row phone">
+        <fieldset class="col-sm-12">
+            <p>Leave this field empty. We do not need your Phone-Number but bots will think so.</p>
+            <p>That way we can eliminate bots!</p>
+            {{ form_row(form.phone) }}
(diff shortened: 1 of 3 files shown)


app/templates/Application/contact.html.twig
```diff
@@ -36,6 +36,13 @@
             {{ form_row(form.comment) }}
         </fieldset>
     </div>
+    <div class="row phone">
+        <fieldset class="col-sm-12">
+            <p>Leave this field empty. We do not need your Phone-Number but bots will think so.</p>
```
iansltx:
```suggestion
            <p>Leave this field empty. We do not need your phone number but bots will think we do.</p>
```

app/templates/Application/contact.html.twig
```diff
@@ -36,6 +36,13 @@
...
+            <p>That way we can eliminate bots!</p>
```
iansltx:
```suggestion
            <p>Thanks for helping us keep this contact form bot-free!</p>
```

PR #787, opened 2019-03-25
Fix legacy links

heiglandreas:
This fixes an issue that @exussum12 discovered.

The links to the Grid and the List-view where both broken due to a trailing slash. As we do not use trailing slashes anywhere I removed possibly trailing slashes via a `trim`-call.

Additionally the Error-Page still showed a link to the legacy-site that by now is not available anymore. So I removed it.

### DIFF
--- a/app/templates/Error/404.html.twig
+++ b/app/templates/Error/404.html.twig
@@ -6,7 +6,7 @@
     <h1 class="title">Page not found</h1>
     <p>Sorry, we can't find the page you tried to visit!</p>
 
-    <p>If you were expecting to see our old site, you can visit it at <a href="http://legacy.joind.in">http://legacy.joind.in</a></p>
+    <p>If you were expecting to see our old site, be informed that it is not available anymore.</p>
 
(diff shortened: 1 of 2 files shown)


app/templates/Error/404.html.twig
```diff
@@ -6,7 +6,7 @@
     <h1 class="title">Page not found</h1>
     <p>Sorry, we can't find the page you tried to visit!</p>
 
-    <p>If you were expecting to see our old site, you can visit it at <a href="http://legacy.joind.in">http://legacy.joind.in</a></p>
+    <p>If you were expecting to see our old site, be informed that it is not available anymore.</p>
```
iansltx:
```suggestion
    <p>If you were expecting to see our old site, it is no longer available as of March 19th, 2019.</p>
```

zghosts:
Might be a separate issue, but shouldn't we also allow for a trailing '/', perhaps redirecting to the same url but without the '/', besides making this change

mcneely:
Ok, finally updated. I had to clear the twig cache on the container.

PR #790, opened 2019-04-08
Don't require the honeypot field, fix honeypot description visibility

iansltx:
I wasn't seeing the description in the contact form.


app/templates/Application/contact.html.twig
```diff
@@ -36,19 +36,19 @@
...
+        <div class="col-sm-12">
+            <input type="submit" class="btn btn-primary pull-right" value="Submit" />
+        </div>
+    </div>
     <div class="row phone">
         <fieldset class="col-sm-12">
             <p>Leave this field empty. We do not need your phone number but bots will think we do.</p>
             <p>Thanks for helping us keep this contact form bot-free!</p>
+        </fieldset>
+        <fieldset class="col-sm-12">
```
mcneely: The description and phone # were supposed to be intentionally hidden by CSS IIRC as we don't require a phone. If it's filled out then it's a bot who scraped the page. Not sure why phone is still showing up on the form.

heiglandreas:
The honeypot field needs to be required but needs to be empty to fullfil its purpose. And it should not be visible to the end user at all. That row should be completely hidden...

iansltx:
@heiglandreas The Required attr disallows form submission via HTML5 constraints (aka any modern browser) unless the field has something in it. But we fail the field if it's populated server-side. As deployed right now, this is the definition of a Catch-22 unless you disable HTML5 form validation. Which, as it turns out, is much easier to do with a scraper/bot than a legitimate user's web browser :upside_down_face:

heiglandreas:
Well, It makes sense then to remove the `required` attribute then from the HTML. Still the field needs to be present with no value on the server side. I might add an additional check besides the usual `form->validate()`....

Having that check within that would be easier but well...

mcneely:
It might work better if we provide a default phone number and expect it back "867-5309" for instance. If it's not there the honeypot fails.

svpernova09:
Can we get a quick recap/summary of tasks to be done to resolve this?

svpernova09:
Closing this as no longer needed. If needed lets discuss.

PR #803, opened 2019-04-29
Platform Image Update

mcneely:
Platform switched their logo to an embedded svg. Swapped out image with actual svg.


svpernova09:
Can someone kick off a deployment so we can get this fixed on the site please.

PR #809, opened 2019-05-03
Add Import Talk from CSV Feature for Event Organizers

svpernova09:
* #586
* JOINDIN-527

Example CSV contents:
```
Sample Talk,Description of Talk,Joe Ferguson,07/19/2019,02:00 PM,English - US,Talk,Track 1
Sample Workshop,Description of Talk 2,Joe Ferguson,07/19/2019,09:00 AM,English - US,Workshop,Track 2
```


app/src/Event/EventController.php
```diff
@@ -36,6 +36,7 @@ protected function defineRoutes(Slim $app)
         $app->get('/event', array($this, 'index'))->name("events-index");
         $app->get('/event/pending', array($this, 'pending'))->name("events-pending");
         $app->map('/event/submit', array($this, 'submit'))->via('GET', 'POST')->name('event-submit');
+        $app->map('/event/:friendly_name/import', array($this, 'eventImportCsv'))->via('GET', 'POST')->name("event-import-csv");
```
heiglandreas: Can we rename that endpoint to "importcsv" (or "import/scv")? That will allow us to add an XML-Import or any other kind of import later without having to change the API.

Alternatively the endpoint should check the fileformat and decide based on that what kind of import to use. But that'd IMO require the method to be renamed to "eventImport"

app/src/Event/EventController.php
```diff
@@ -1149,6 +1150,74 @@ public function editTracks($friendly_name)
...
+
+                    while (!feof($handle)) {
+                        $talk = fgetcsv($handle);
+                        $date_start = new \DateTime();
+                        $date_start->setTimestamp(
+                            strtotime($talk[3].' '.$talk[4])
+                        );
+                        $date_end = new \DateTime();
+                        $date_end->setTimestamp(
+                            strtotime($date_start->format($date_end::ATOM).' plus '.$talk[5].' minutes')
```
heiglandreas: Why the time-magic here? Assuming that the CSV-File contains the fields "date", "starttime" and "duration" you could also do this:

```php
$date_start = new DateTimeImmutable($talk[3] . ' ' . $talk[4]);
$date_end = $date_start->add(new DateInterval('PT' . $talk[5] . 'M'));
```

Additionally I'm missing the Timezone-Information. As far as I can tell this will then be imported using the local timezone which might lead to unexpected results. 

So perhaps something like this would help here:

```php
$date_start = new DateTimeImmutable($talk[3] . ' ' . $talk[4], $event->getFullTimezone());
```

app/src/Event/EventController.php
```diff
@@ -1149,6 +1150,74 @@ public function editTracks($friendly_name)
...
+
+                        $talk_data = [
+                            'talk_title' => $talk[0],
+                            'talk_description' => $talk[1],
+                            'type' => $talk[7],
+                            'track' => $talk[8],
+                            'language' => $talk[6],
+                            'start_date' => $date_start->format('c'),
+                            'speakers' => [$talk[2]],
+
```
heiglandreas: Shouldn't either the talk duration or the end-time of the talk be part of this data-array? If neither is used here why did we calculate it in the first place?

app/src/Event/EventController.php
heiglandreas: The timezone seems to be irrelevant here as the DateTimeInterface will be "reduced" to `Y-m-d H:i` almost immediately in `TalkApi::addTalk`. Still it might help with reducing errors due to wrongly set default-timezone.

app/src/Event/EventController.php
heiglandreas: Especially when the duration will be set by default to 60 minutes...

app/src/Event/Constraint/ValidCsvFile.php
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @Annotation
+ */
+class ValidCsvFile extends Constraint
+{
+    public $groupname;
+    public $keyname;
+
+    public $message = 'The CSV must be formatted properly';
```
heiglandreas: Will this class check in the future whether there are the right fields in the right order?

app/src/Event/EventController.php
svpernova09: I have no plans to support anything but CSV importing. If someone wanted to add XML importing in the future, I'd push back and not want to maintain 2 import formats.

app/src/Event/EventController.php
svpernova09: Like you stated, all the times are being handled by the Talkapi endpoint and methods. As long as my CSV has the talks in the event's timezone it'll work out as expected.

app/src/Event/Constraint/ValidCsvFile.php
svpernova09: It can do whatever we want it to in the future.

app/src/Event/EventController.php
heiglandreas: It'll not necessarily work out as expected. But that will only happen in so weird special edge-cases that I'm with you here that the timezone-information can be skipped here.

Should weird things happen they need to be addressed later via the web-UI

app/src/Event/EventController.php
svpernova09: Good catch, yes I dropped the duration field during debugging. Added it back.

svpernova09:
Ping @joindin/leadership-team

app/src/Event/EventController.php
```diff
@@ -1149,6 +1150,74 @@ public function editTracks($friendly_name)
...
+        $config = $this->application->config('oauth');
+        $request = $this->application->request();
+
+        /** @var FormFactoryInterface $factory */
+        $factory = $this->application->formFactory;
+        $form    = $factory->create(new EventImportFormType());
+
+        //use EventFormImportType to validate the CSV is valid
+        //We possibly want to do some extensive data checking
+        //BEFORE we blindly throw data from here to the API
```
zghosts: "possibly" is an understatement here, we need to add validation here as there is no validation being done in the api, so it must be done here

svpernova09: :(

svpernova09:
Data coming in from the CSV is now sanitized before sending to the API

app/src/Event/EventController.php
```diff
@@ -1149,6 +1150,74 @@ public function editTracks($friendly_name)
...
+                    $event = $eventApi->getByFriendlyUrl($eventSlug);
+                    $handle = fopen($_FILES['event_import']['tmp_name']['csv_file'], "r");
+
+                    while (!feof($handle)) {
+                        $talk = fgetcsv($handle);
+                        $date_start = new \DateTime();
+                        $date_start->setTimestamp(
+                            strtotime(filter_var($talk[3], FILTER_SANITIZE_STRING).' '.filter_var($talk[4], FILTER_SANITIZE_STRING))
+                        );
+                        $date_end = new \DateTime();
```
zghosts: $date_end isn't used so it can be removed

app/src/Event/EventController.php
```diff
@@ -1149,6 +1150,74 @@ public function editTracks($friendly_name)
...
+            try {
+                if (isset($_FILES['event_import']['error']['csv_file'])
+                    && $_FILES['event_import']['error']['csv_file'] == UPLOAD_ERR_OK) {
+                    $eventApi = $this->getEventApi();
+                    $event = $eventApi->getByFriendlyUrl($eventSlug);
+                    $handle = fopen($_FILES['event_import']['tmp_name']['csv_file'], "r");
+
+                    while (!feof($handle)) {
+                        $talk = fgetcsv($handle);
+                        $date_start = new \DateTime();
```
zghosts: There is no need to use strtotime and timestamps here, this does the same thing:

$date_start = new \DateTimeImmutable(filter_var($talk[3], FILTER_SANITIZE_STRING).' '.filter_var($talk[4], FILTER_SANITIZE_STRING));

app/src/Event/EventController.php
```diff
@@ -1149,6 +1150,74 @@ public function editTracks($friendly_name)
...
+                            strtotime($date_start->format($date_end::ATOM).' plus '.filter_var($talk[5], FILTER_SANITIZE_STRING).' minutes')
+                        );
+
+                        $talk_data = [
+                            'talk_title' => filter_var($talk[0], FILTER_SANITIZE_STRING),
+                            'talk_description' => filter_var($talk[1], FILTER_SANITIZE_STRING),
+                            'type' => filter_var($talk[7], FILTER_SANITIZE_STRING),
+                            'track' => filter_var($talk[8], FILTER_SANITIZE_STRING),
+                            'language' => filter_var($talk[6], FILTER_SANITIZE_STRING),
+                            'start_date' => $date_start->format('c'),
```
zghosts: format('c') adds a timezone, unnecessarily shifting the time
```
'start_date' => $date_start->format('Y-m-d H:i'),
```
works

svpernova09:
Updated based on @zghosts comments. Thanks!

zghosts:
Maybe also add a description of the csv fileformat to the template when uploading, so the uploader doesn't have to guess the format

svpernova09:
> Maybe also add a description of the csv fileformat to the template when uploading, so the uploader doesn't have to guess the format

Done! Should be good to go now.

zghosts:
I'm working on a format ValidCsvFileValidator in order to make the ValidCsvFile Constraint actually work so the entire file's structure can be validated before starting the import. Should be able to finish it this evening.
I would like to see some validation improvements before we deploy this to production as the implementation is somewhat brittle if a record somewhere halfway isn't formatted correctly. If you want I can approve this PR now and add the validation in a separate PR, that way it's easier for others to improve on this.

zghosts:
I might be nitpicking here but from a UX point of view, maybe adding some feedback after importing might also be a nice, telling the user what has been imported, instead of reshowing the form without any feedback.

If you want to ship this as is and improve the UX at a later stage I'm sortof fine with that, as long as it will be done.

svpernova09:
I would prefer to ship this and let someone who knows what they’re doing in regard to UX adjust things later. I don’t want to get bogged down waiting for “perfect”.

--
- Joe Ferguson
JoeFerguson.me
osmihelp.org

svpernova09:
Let’s make sure we open an issue to describe what needs to be done UX Wise. I’ll make sure I have time to test the validation pr as soon as it comes across.

--
- Joe Ferguson
JoeFerguson.me
osmihelp.org

PR #249, opened 2015-09-09
JOINDIN-575 [WIP] Validate URLs before at least displaying them, better before storing

shaunmza:
PR for code / functionality review see https://joindin.jira.com/browse/JOINDIN-575
Added changes to try and resolve url's on saving an event.
Short url's are followed to the final url, if the url does not resolve the save fails

PR #824, opened 2019-06-30
JOINDIN-575 Added changes to try and resolve url's on saving an event.

heiglandreas:
Short url's are followed to the final url, if the url does not resolve the save fails


heiglandreas:
This PR replaces #249.

zghosts:
I want to say, could you please add some tests? At least for the constraints.

heiglandreas:
I can. As this is a PR from 2015 it's unfair tonask the original contributor, so I'll see what I can do tomorrow during a train ride 😁

zghosts:
2015? That’s old, asking the original contributer would indeed be unfair, do what you can. tests for the heavy lifting parts would be great

heiglandreas:
Additional testing added. And some minor refactoring as well...

PR #828, opened 2019-07-08
Allow null-values for URL validation

### LINKED ISSUE #827: Unable to Add Event Without CFP Link
When adding an event through the UI, the CFP URL is optional. However, not including one results in an error when adding the event.

As our meetup events don't have a CFP, it would be ideal to have this field continue to be optional

Error on submission:

`Fatal error: Uncaught TypeError: Argument 1 passed to Form\Exception\UnresolvableUrl::__construct() must be of the type string, null given, called in /app/joindin-web2/app/src/Form/Shared/UrlResolver.php on line 35 and defined in /app/joindin-web2/app/src/Form/Exception/UnresolvableUrl.php:11 Stack trace: #0 /app/joindin-web2/app/src/Form/Shared/UrlResolver.php(35): Form\Exception\UnresolvableUrl->__construct(NULL, 0) #1 /app/joindin-web2/app/src/Form/Listener/GetResolvedUrlListener.php(27): Form\Shared\UrlResolver->resolve(NULL) #2 [internal function]: Form\Listener\GetResolvedUrlListener->onSubmit(Object(Symfony\Component\Form\FormEvent), 'form.bind', Object(Symfony\Component\EventDispatcher\EventDispatcher)) #3 /app/joindin-web2/vendor/symfony/event-dispatcher/EventDispatcher.php(174): call_user_func(Array, Object(Symfony\Component\Form\FormEvent), 'form.bind', Object(Symfony\Component\EventDispatcher\EventDispatcher)) #4 /app/joindin-web2/vendor/symfony/event-dispatcher/EventDispatcher.php(43): Symfony\Component\Event in /app/joindin-web2/app/src/Form/Exception/UnresolvableUrl.php on line 11`

heiglandreas:
This allows URLs that will be validated to also have a value of NULL instead of only allowing an empty string. Due to the way the values are passed into the validation a non-existent value will become a NULL-value and therefore failed before this fix.

Fixes #827

### DIFF
--- a/app/src/Form/Constraint/UrlResolverConstraintValidator.php
+++ b/app/src/Form/Constraint/UrlResolverConstraintValidator.php
@@ -15,6 +15,10 @@ class UrlResolverConstraintValidator extends ConstraintValidator
             return;
         }
 
+        if ($value === null) {
+            return;
+        }
+
         $resolver = new UrlResolver();
         try {
             $resolver->resolve($value);
--- a/app/src/Form/Listener/GetResolvedUrlListener.php
+++ b/app/src/Form/Listener/GetResolvedUrlListener.php
@@ -37,6 +37,6 @@ class GetResolvedUrlListener implements EventSubscriberInterface
 
     public static function getSubscribedEvents()
     {
-        return array(FormEvents::SUBMIT => 'onSubmit');
+        return [FormEvents::SUBMIT => 'onSubmit'];
     }
 }
--- a/app/src/Form/Listener/GetResolvedUrlListener.php
+++ b/app/src/Form/Listener/GetResolvedUrlListener.php
@@ -13,7 +13,6 @@ use Form\Shared\UrlResolver;
  */
 class GetResolvedUrlListener implements EventSubscriberInterface
 {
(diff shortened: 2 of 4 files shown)


app/src/Form/Constraint/UrlResolverConstraintValidator.php
```diff
@@ -15,6 +15,10 @@ public function validate($value, Constraint $constraint)
             return;
         }
 
+        if ($value === null) {
```
mcneely: wouldn't using an if (empty($value)) on the previous if check work better?

heiglandreas: `empty()` here would allow all values that cast to `false` like integers, floats and booleans also. I specifically want to only handle an empty string as well as the `null`   value. Even the `null` value is a concession to the way the form is handled internally where instead of an empty string a `null` value is passed...

mcneely: oh, I see now. It would have helped if I actually had read the description of the pr.

PR #852, opened 2019-10-04
Update footer w/ links

svpernova09:
Screen Shot 2019-10-03 at 19 14 50


svpernova09:
> Mind adding text descriptions for the GH/blog/Twitter links? Otherwise we have a bit of an a11y issue here.

Good call, added titles.

iansltx:
Hm, test appears to be failing for some reason.

svpernova09:
> Hm, test appears to be failing for some reason.

I've had to restart test runners several times. Not sure the reason (I should probably check) but a restart build usually passes.

Issue #806, opened 2019-05-02
Older Anonymous Comments - Now half-aren't


EliW:
Look here: https://joind.in/event/phptek-2018/i-community-and-so-can-you#comment-e66dc3

A comment that was left under the old 'Make This Anonymous' feature.   It's currently showing the user's headshot, which kinda breaks the whole anonymous piece.

dstockto:
@EliW Do you have any examples of truly anonymous comments (like from back-in-the-day when you could log in and mark a comment as anonymous)? The one mentioned is, as far as I can tell, definitely associated to a user, but that user has no username, but does have an email address. I've been trying to figure out how comments used to be marked anonymous, but I think this one is different.

e3betht:
@dstockto - Is this what you are looking for?
https://joind.in/event/madison-php-conference-2014/make-your-code-behave#comment-f3cf66

dstockto:
@e3betht  - As far as I can see, there's no link to the user who made the anonymous comment on that example, so it seems to be truly anonymous. If there was/is a link, I don't know how to find it. The example that Eli sent seems to have had a user who then didn't have a username, but did have an email that was mapped to a Gravatar. If the email wasn't on Gravatar then I don't know that there's any easy way to find those from the web interface.

EliW:
@e3betht - I can look for one, but not now (php[world] next week) ...

However I would assume that you'll need to look in the database to find this info.   As during "era 2" of Joind.in ... when anonymous cowards were removed, but you could choose to make a comment anonymous ... there obviously needed to be a flag in the database of some sort to mark that comment as such.

(Though the 2.0 refactor may have blown all that away?  Which might explain what you are seeing?)

EliW:
What's interesting I just realized, is that that example I gave was 2018 ... there weren't any anonymous comments in 2018.   So obviously that one, in particular, was just a bug of some sort.

e3betht:
@EliW Anonymous comments were not fully turned off until January of 2019, so there were anonymous comments in 2018.

heiglandreas:
Until beginning of this year the "old" website was still online and it might still have been possible to give anonymous comments via that interface. And we usually didn't remove fields from the database due to the rewrite. We extended the DB but - especially because of the legacy site - we very rarely removed stuff to not break anything...

dstockto:
[...]

PR #856, opened 2019-10-13
Show anonymous gravatar for users with no username

dstockto:
Fix for #806

The comment in #806 is not really anonymous, at least not in any way I can see. The user that's on there has no username. So I'm not sure if that's something to do with account deletion, or whatever... But it has an email address. This fix should put the anonymous gravatar in place if the user who made a comment doesn't have a username. 

I haven't been able to find an example of any logged-in, but checked as anonymous that the site used to allow. If anyone has an example of one of those, it'd be pretty helpful.

### DIFF
--- a/app/templates/_common/comment.html.twig
+++ b/app/templates/_common/comment.html.twig
@@ -17,7 +17,11 @@
                 <div class="row">
                     <h3>
                         <div class="avatar pull-right">
-                            <img src="{{ gravatar(comment.getGravatarHash) }}">
+                            {% if comment.getUserDisplayName %}
+                                <img src="{{ gravatar(comment.getGravatarHash) }}">
+                            {%  else %}
+                                <img src="https://secure.gravatar.com/avatar/294de3557d9d00b3d2d8a1e6aab028cf?d=mm&s=40">
+                            {% endif %}
                         </div>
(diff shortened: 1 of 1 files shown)


app/templates/_common/comment.html.twig
```diff
@@ -17,7 +17,11 @@
                 <div class="row">
                     <h3>
                         <div class="avatar pull-right">
-                            <img src="{{ gravatar(comment.getGravatarHash) }}">
+                            {% if comment.getUserDisplayName %}
+                                <img src="{{ gravatar(comment.getGravatarHash) }}">
+                            {%  else %}
+                                <img src="https://secure.gravatar.com/avatar/294de3557d9d00b3d2d8a1e6aab028cf?d=mm&s=40">
```
heiglandreas: Shouldn't we use the way that is described at https://en.gravatar.com/site/implement/images/ under "Default Image"? 

Instead of providing a hash to an image that is nowhere documented where it leads to? what email address is that? who's in charge of that account?

dstockto: That's the md5 hash of "anonymous"

dstockto: Updated to all 0s and default forced to mystery person

PR #859, opened 2019-10-18
Extract profile navigation to template, Add pending field to event data

svpernova09:
2 tasks accomplished here

1. Clean up profile navigation by extracting to it's own include template
1. Add method to return pending status for displaying to user an event they host is pending

### DIFF
--- a/app/templates/User/profile-events.html.twig
+++ b/app/templates/User/profile-events.html.twig
@@ -3,7 +3,7 @@
 {% block title %}Events {{ type }} by {{ thisUser.getFullname }} - Joind.in{% endblock %}
 
 {% block body %}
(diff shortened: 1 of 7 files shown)


app/templates/User/profile-events.html.twig
```diff
@@ -3,40 +3,7 @@
...
-                            </a>
-                        </p>
-                    </section>
-
-                </div>
-            </div>
-        </div>
-    </section>
-
+    {% include 'User/profile-navigation.html.twig' %}adsfasdfasdfa
```
mcneely: Is this an oops?

svpernova09: rofl, yep!

svpernova09: Fixed!

PR #875, opened 2019-10-23
Refactor UrlResolverConstraintValidatorTest

mishavantol:
- assert addViolation is not called in testValidateWorks.
- change deprecated interface
- don't assert null on a method with void result

This test was failing the first time I ran all tests. Actually making the curl call is probably not a good idea because it gives random false negatives. We could also just check if the resolve method is called instead of making the actual call?


svpernova09:
Looks like a test broke with this change:

```
1) Tests\Form\Constraint\UrlResolverConstraintValidatorTest::testValidateWorks with data set #0 ('http://apple.de')

Symfony\Component\Validator\ExecutionContextInterface::addViolation('Url provided does not resolve.', Array (), null, null, null) was not expected to be called.
```

mishavantol:
http://apple.de is giving an connectionError every once in a while at the moment. I'll update the test tomorrow so that it doesn't make the actual curl call anymore. It looks like if we're testing the resolver here, but actually we're just testing if the internet and the apple.de webserver are working as expected.

mishavantol:
There are more tests making (the same) external requests. I was planning on making separate PR's for them. But I'll add them to this PR to prevent the build from failing randomly.

mishavantol:
Done, the UrlResolver is mocked and doesn't curl anymore in the tests. This makes the test suite stable (and a bit faster, ~5s vs ~7s). Now I can finally pick up some other issues. (hacktober FTW!) 😄

Issue #773, opened 2019-03-05
Missing X-Content-Type-Options Header


aenglander:
X-Content-Type-Options stops a browser from trying to MIME-sniff the content type and forces it to stick with the declared content-type. The only valid value for this header is "X-Content-Type-Options: nosniff".

Issue #772, opened 2019-03-05
Missing X-XSS-Protection Header


aenglander:
X-XSS-Protection sets the configuration for the cross-site scripting filter built into most browsers. Recommended value "X-XSS-Protection: 1; mode=block".

Issue #771, opened 2019-03-05
Missing X-Frame-Options Header


aenglander:
X-Frame-Options tells the browser whether you want to allow your site to be framed or not. By preventing a browser from framing your site you can defend against attacks like clickjacking. Recommended value "X-Frame-Options: SAMEORIGIN".

PR #880, opened 2019-10-24
Added extra security headers #771 #772 #773

mslepko:
Added extra security headers per requirements in #771 #772 #773

### DIFF
--- a/web/.htaccess
+++ b/web/.htaccess
@@ -8,6 +8,18 @@ RewriteEngine On
 #
 # RewriteBase /
 
+# Add Security Headers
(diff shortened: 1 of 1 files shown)


iansltx:
We're actually using nginx to serve this so these need to be handled at the app level.

mslepko:
Ok, I saw .htaccess file and thought it would be the best place to add it.
Should I add it in /web/index.php? Something like
```
header('X-Frame-Options: SAMEORIGIN);
```

svpernova09:
@mslepko yes I believe so.

PR #882, opened 2019-10-27
Adjust homepage layout to 2 major columns

svpernova09:
Here's a basic tweak of the Hompage with the specific goals of showing upcoming events AND recent events. This primarily targets 2 users: Users coming to rate talks they've just seen, or view talks at upcoming events.

Open to discussion if anyone has suggestions.

## Logged out Preview:

Screenshot_2019-10-26 Joind in

## Logged In Preview:
Screenshot_2019-10-26 Joind in(1)


iansltx:
How does this look on mobile? Afraid that the recent column would end up with a random amount of scrolling. Got some ideas to mitigate (would end up looking like the old Hot Events, but with a tweaked algorithm), but will probably be Tuesday before I can sit down and lay 'em out.

svpernova09:
Here's what it looks like on mobile, 2 VERY slim columns.

Screenshot_2019-10-27 Joind in

svpernova09:
So this is 100% up for discussion. This introduces another set of talks to display on the home page bringing the total number to 3 talk sections: Upcoming Events, Past Events, and CFP Events. This change was made due to feedback during php[world] 2019 as members had a hard time finding the event due to the cutoff date / timezone being in UTC. 

I don't want to get bogged down into the weeds.

If we want to go back to hot events and modify the algorithm, I'd like to get a PR worked up to see that. Otherwise, we should figure out which types of events users are going to most often and focus there.

PR #886, opened 2019-10-27
 Use rowspan to properly visualize double-length talks in grid view Fix  #879

### LINKED ISSUE #879: Use rowspan to properly visualize double-length talks in grid view
e.g. https://joind.in/event/phpworld-2019/schedule/grid

This would be useful for conferences that have either double-length workshops or half-length uncon talks running alongside a standard talk length.

Below is an example that I got my adding rowspans and deleting cells manually:

Double-length talk with rowspans

lfbittencourt:
Now the grid is using rowspan attribute to "properly visualize double-length talks in grid view". Not only double-length, actually, but any N-length talks :-)


svpernova09:
#879

PR #887, opened 2019-10-27
Improve local dev

lfbittencourt:
I followed `README-local-dev.md` instructions to run the project for the first, however several static assets as JavaScript files were missing since they were not listed in `web/index.php` static file extensions. I replaced that part with a regular expression and added all missing extensions that I found.

This PR also includes a little extra instruction about how to use Docker to run a Redis instance.

### DIFF
--- a/README-local-dev.md
+++ b/README-local-dev.md
@@ -1,6 +1,6 @@
 # Quick start with PHP dev server
 
-The web2 site will run happily under the [PHP development server](http://php.net/manual/en/features.commandline.webserver.php), but you'll need to have a [Redis](http://redis.io) server running on localhost:6379. You can easily do this with Docker typing `docker run --name some-redis -d -p 6379:6379 redis`.
+The web2 site will run happily under the [PHP development server](http://php.net/manual/en/features.commandline.webserver.php), but you'll need to have a [Redis](http://redis.io) server running on localhost:6379. You can do this with Docker typing `docker run --name some-redis -d -p 6379:6379 redis`.
 
 You'll also need to configure a suitable [Joind.in API](https://github.com/joindin/joindin-api/); you could use the [live Joind.in API](https://api.joind.in/) or if you're developing against a local version, reference it in config/config.php.  Note that the API will also run under the built-in webserver, but will need to be on a different listening port.
 
--- a/README-local-dev.md
+++ b/README-local-dev.md
@@ -2,7 +2,7 @@
 
 The web2 site will run happily under the [PHP development server](http://php.net/manual/en/features.commandline.webserver.php), but you'll need to have a [Redis](http://redis.io) server running on localhost:6379. You can do this with Docker typing `docker run --name some-redis -d -p 6379:6379 redis`.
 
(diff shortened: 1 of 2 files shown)


svpernova09:
I don't *think* we want to make changes to `index.php` to make things in local dev work. Will defer to @dstockto or @ericpoe

lfbittencourt:
@svpernova09 The static files check already exists, so it's a simple fix IMO. Let me know if I can do any improvements on this!

README-local-dev.md
```diff
@@ -1,6 +1,8 @@
 # Quick start with PHP dev server
 
-The web2 site will run happily under the [PHP development server](http://php.net/manual/en/features.commandline.webserver.php), but you'll need to have a [Redis](http://redis.io) server running on localhost:6379.  You'll also need to configure a suitable [Joind.in API](https://github.com/joindin/joindin-api/); you could use the [live Joind.in API](https://api.joind.in/) or if you're developing against a local version, reference it in config/config.php.  Note that the API will also run under the built-in webserver, but will need to be on a different listening port.
+The web2 site will run happily under the [PHP development server](http://php.net/manual/en/features.commandline.webserver.php), but you'll need to have a [Redis](http://redis.io) server running on localhost:6379. You can easily do this with Docker typing `docker run --name some-redis -d -p 6379:6379 redis`.
```
ericpoe: Let's try to avoid words like "easily" and "simply" since these are subjective and might not be as "easy" or "simple" to the end user as it is to the author.

```suggestion
The web2 site will run happily under the [PHP development server](http://php.net/manual/en/features.commandline.webserver.php), but you'll need to have a [Redis](http://redis.io) server running on localhost:6379. You can do this with Docker typing `docker run --name some-redis -d -p 6379:6379 redis`.
```

README-local-dev.md
```diff
@@ -1,6 +1,8 @@
...
+
+You'll also need to configure a suitable [Joind.in API](https://github.com/joindin/joindin-api/); you could use the [live Joind.in API](https://api.joind.in/) or if you're developing against a local version, reference it in config/config.php.  Note that the API will also run under the built-in webserver, but will need to be on a different listening port.
```
ericpoe: I don't think that we want anyone to be developing against the live API. It would be too easy to accidentally submit bogus data and tests would also submit bogus data. Using the local-dev API would be better for dev work.

lfbittencourt:
@ericpoe Thanks! I followed your advice and updated `README-local-dev.md` file. Note, however, that it doesn't address @svpernova09 question about `index.php` static assets checking.

I think it was wrong mix both things on the same PR, but do you think we can merge it now? I can close this one and submit two separated PRs if you prefer.

ericpoe:
Thanks for updating the instructions, that will certainly help the docker users.

Yes, this PR could have been two separate PRs since it fixes two completely separate problems, which makes reviewing and merging a lot easier.

ericpoe:
I am trying to recreate the problem that the code portion of this PR is fixing. Where are you seeing missing assets? Is this in a specific tool? In dev vs production?

PR #888, opened 2019-10-27
 Add talk end time to talk detail view Fix #872

### LINKED ISSUE #872: Add talk end time to talk detail view
This matters a bit more for conferences that have split-duration tracks (e.g. php[world]), where talks that start at the same time don't necessarily end at the same time. We store the information but don't display it on the individual talk info page.

lfbittencourt:
Add talk end time to talk detail view.


svpernova09:
#872

PR #889, opened 2019-10-27
Fix #832

### LINKED ISSUE #832: Encoding error - Talk Titles
**Bug:**
If you put an HTML Encoded character in a talk title.  On the 'grid' page it shows the encoding.  On the talk page, the title is passed raw and therefore the decoded character is shown.

**Replicate:**
Create or edit a talk title.  Put &#39; in it.  Save the talk.   View the grid page, and you will see &#39; -- View the talk page, and the title will show a '

**Expected Behavior:**
When entering &#39; in a title, EITHER:
1) That HTML Encoding will be automatically decoded and saved as such.
2) That &#39; should show up on the talk title page itself.

**Real Bug:**
I assume that the &#39; is being stored in the database, and that the Talk Title isn't being html_specialchars() before being output on the talk page.   Which means that someone could theoretically inject things into talk titles.

lfbittencourt:
Talk title is shown consistently in list and grid views, however there was a explicit raw filter in the talk info page. I don't know why it was there, but now it's gone.

### DIFF
--- a/app/templates/Talk/_common/talk_header.html.twig
+++ b/app/templates/Talk/_common/talk_header.html.twig
@@ -2,7 +2,7 @@
     <div class="row event">
         <div class="col-xs-12 col-sm-12 title">
             <h1>
-                {{ talk.getTitle | raw }}
+                {{ talk.getTitle }}
                 {% if user %}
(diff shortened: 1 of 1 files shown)


svpernova09:
#832

PR #890, opened 2019-10-28
Use event timezone on schedules

svpernova09:


iansltx:
Believe that the problem this was attempting to resolve got solved by https://github.com/joindin/joindin-api/pull/761. The issue also affected the talk edit page, so if we want to go forward with a web2-side fix we'd want to tweak that as well. Or just close this.

PR #899, opened 2020-01-24
Bail out more carefully when Slim asks the app to halt

iansltx:
e.g. when redirecting as a result of the trailing slashes middleware; this should allow the normal app life cycle to operate on a redirect or the like rather than throwing an exception, turning a redirect into
a 500.

Looked at how Slim 2's run() method works and this seems to be the best bet to resolve.


heiglandreas:
Looks good from my point.

iansltx:
Whoops, had some stray bits in there, and this was out of date. Let's try that again.

PR #911, opened 2023-05-20
Set the report link type to button

svpernova09:
Working to debug why "Report comment" seems to be broken.

This PR is the next step. An attempt to set a HREF to a button.

Not able to reproduce locally but not in production.

Works locally:

image

But not in production:

image

Slack convo Debugging w/ @iansltx so far:

```
Joe: I'm testing if https://www.google.com/search?client=firefox-b-1-d&q=window.__cfRLUnblockHandlers is causing our "Report a Comment" to throw "method not found". Reporting comments is fine for me in local dev. But not in prod. Inspecting the HTML shows that __cfRL being injected into our buttons.

Ian: Got it. I can turn off Rocket Loader...and done. lmk if that fixes it

Joe: Nope, no luck. I guess we can turn it back on. I confirmed I didn't see that code in the HTML as I did before, but no change. Thanks for the help. 
```

### DIFF
--- a/app/src/Apikey/ApikeyDeleteFormType.php
+++ b/app/src/Apikey/ApikeyDeleteFormType.php
@@ -12,7 +12,6 @@ use Event\EventEntity;
  */
 class ApikeyDeleteFormType extends AbstractType
 {
-
     /**
      * Returns the name of this form type.
      *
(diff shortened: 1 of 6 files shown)


iansltx (approved):
Not sure why this would fix, but let's see what happens.

PR #913, opened 2023-05-23
Enable GitHub Actions

### LINKED ISSUE #912: GitHub Actions?
[jenkins.joind.in](https://jenkins.joind.in/) is no longer in operation.  Should I set up GitHub Actions for the project?

I'm interested in helping and having actions in place will make me feel better about submitting a PR.  :-)

warren5236:
This create a GitHub Actions file that's a combination of the GitHub Action file in the api repository and the actions that were being performed in the build.sh file.  `vendor/bin/security-checker` doesn't work anymore so I didn't move it over. I still need to get PHPStan and composer-require-checker installed and running which I will do in another PR unless you want it here. The API repository is uploading results to codecov but the project was deleted do you want that here?

Closes #912


iansltx:
We can go ahead without codecov on this repo. We can add code coverage reporting in later.

Given the above, looks like we're a readme tweak or two away from being merge-able, on first glance.

iansltx:
Also, let's run build.sh directly in Actions rather than deleting it, for a little more portability.

iansltx:
Finally, if you don't mind squashing fix commits down for a cleaner history, would be appreciated.

Issue #896, opened 2019-11-07
Make track selection optional when tracks are available


iansltx:
Social events, panels, etc. may not map to a specific track, but we require selection of a track on every talk anyway, at least if tracks are specified at all. Track association should be optional in these cases, and the API already treats tracks as optional.

PR #916, opened 2023-09-29
Added track names to talk detail view

### LINKED ISSUE #871: Add track name to talk detail view
Right now it's only visible in the list view. Should be a quick add, provided we don't link-ify the name (there's nowhere to link to at this point anyway).

warren5236:
I know there is an issue (#896) to make the track optional so I make this change assuming that would get taken care of soon. 

I'm trying to contribute this as part of hacktoberfest but I'm not able to select the label.

This closed #871


afilina:

heiglandreas (approved):

PR #917, opened 2023-10-06
Make the track selection option on the talk form.

### LINKED ISSUE #896: Make track selection optional when tracks are available

warren5236:
This PR makes the "track" selection optional on the talk form.

Closes #896


afilina:

Issue #690, opened 2016-09-10
JOINDIN-701: If you login on /not-allowed, you go back to that page


ghost:
Jira issue originally created by user liam-wiltshire:

If you try to perform an action that you should be able to, but if your session has been removed or you've been logged out, you are redirected (correctly to /not-allowed). However, if you then re-login on that page (using username/password), you are redirected back to /not-allowed, which make it look like whatever action you were trying to perform is still disallowed.

I would suggest that when we redirect someone to /not-allowed we keep a track of their previous page, and then re-route them back to that page if they login.

PR #919, opened 2023-10-06
Redirect check so redirect to /not-allowed sends the user back to /

warren5236:
Was looking at #690 and found this as a quick fix if it works for you all.


afilina:

